PR #1, opened 2017-11-21
Initial gitconfigs

sarcasticadmin:
Adding more files that should be ignored and enabled `git lfs` for `.pdfs`.

@irabinovitch @hriday let me know what you think!


sarcasticadmin:
thanks @irabinovitch!

PR #9, opened 2018-02-09
More config file progress.

owendelong:
## Description of PR
Brief description of the PR

## Previous Behavior

## New Behavior

## Tests
How was this PR tested?

### DIFF
--- a/switch-configuration/README.md
+++ b/switch-configuration/README.md
@@ -5,8 +5,38 @@ configurations, tooling and scripts for the Juniper Switches and Routers running
 PERL 5
 
 # Configuration Files (User Servicable)
+All files have the following features unless expressly stated otherwise:
+
+```
++	All text after // is treated as a comment and ignored by the parser.
++	Any line ending in a space followed by a backslash (matches /\s\\$/)
+	will result in the next line being treated as a continuation of the
+	existing line. Whitespace at the end of this line and the beginning of the
+	next line will be collapsed to a single space by the parser during joining.
+	The parser will then parse the entire line as normal, including this treatment
+	if the resulting line still ends in a space followed by a backslash.
+
+	e.g.:
+		foo \
+		bar
+
+	is parsed as 'foo bar'
+
+		this \
+		line \
+		is \
+		continued
+
+	is parsed as 'this line is continued'
+```
+## config/defaults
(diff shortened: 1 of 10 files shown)


sarcasticadmin:
@owendelong nice work! you're correctly branching like we had gone over in the demo! 👍 

In regards to your comments:

1. I'm following what you've put together. I'm assuming there's going to be a `perl` script that reads this all in and just spits out the appropriate config?
2. As for the `git` guide its here in `CONTRIBUTING.md`: https://github.com/socallinuxexpo/scale-network/blob/master/CONTRIBUTING.md

owendelong:
Thanks, Robert. Can you approve the review so I can merge?

You are correct that I am planning to write script(s) to pull together configuration files for the defined switches from the data in the files I've been creating.

I'm trying to wrap my head around last year's switch port lists (for example the ballroom C, G IDF switches are quite puzzling so far) and trying to build the smallest number of switch types possible while still meeting our needs. I want each switch from a given class to be nearly identical (differing only in management IP address and certain other parameters).

sarcasticadmin:
Sounds good!

As for the switch types its hard to remember what was special about some of those ballrooms. I'll try to recall what I can so we can work through it tomorrow during the hot stage.

owendelong:
I think I managed to adequately sort it. Thanks for merging. I just pushed v0.0.3. ;-)

Now it's time to start reviewing the content of the files for actual sanity and that it matches expectations for configuration in the real world. I'm trying to make all the switches come as close as possible to looking alike (e.g. all switches that have sign ports have them on the same port numbers).

I am changing the format of the VLAN specification files to add v4 and v6 prefixes. I put that in the documentation, but I haven't gotten to assigning the actual prefixes yet. I will try to do that after I get to Los Angeles tonight.

I want to get on the road soon.

We're probably going to end up testing with manual switch configurations this weekend if we get that far. I'm pedaling as fast as I can. Might ask for help writing some of the scripts (which is one of the reasons I'm pushing so hard to do the config files and formats and documentation first).

I want to be able to have help developing scripts as time gets shorter.

I figure if we have a good design and a good set of template files to start, with well defined requirements for the scripts (added in v0.0.3, btw), we should be able to share that workload if necessary.

PR #26, opened 2018-02-16
[READY] Ansible dynamic inventory

kylerisse:
## Description of PR
Ansible now uses a dynamic inventory which consumes the vlans.d files, switchtypes file, and serverlist.tsv file

## Previous Behavior
static host file and no integration

## New Behavior
dynamically generated inventory, integration with Owen's switch and vlan inventories

## Tests
Locally using vagrant by issuing the _vagrant up_ command within the ansible directory followed by the _vagrant provision_ command to verify idempotence.


sarcasticadmin:
@kylerisse  Ill look through this before the call today

ansible/etc/ansible/.gitignore
```diff
@@ -0,0 +1 @@
+*.retry
```
sarcasticadmin: Thanks for adding this to the ignore :+1:

ansible/inventory.py
```diff
@@ -0,0 +1,146 @@
...
+# Dynamic inventory script used to slurp in various
+# SCaLE specific text files to produce a sane inventory to ansible
+
+import os
+import re
+
+# constants with locations to files or directories within the repo
+vlansddir = "../switch-configuration/config/vlans.d/"
+switchesfile = "../switch-configuration/config/switchtypes"
+serverfile = "../facts/servers/serverlist.tsv"
```
sarcasticadmin: I'm digging this idea!

ansible/todo
```diff
@@ -0,0 +1,15 @@
+todo:
```
sarcasticadmin: Ill move some of this to issues since so we can get better visibility into what we'll need to talk over the next couple of weeks

kylerisse:
roger that on the typo @owendelong , thanks, fixed in local work

per call:
ntp, dns, dhcp, and loghost will all be the two core servers listed below (::5 & ::6). The plan for dhcp is to have the "building local" server hand out ipv4 to it's building. loghost can be pointed at the local building core server, with server2 forwarding to server1. 

from serverlist.tsv:
```
//name	mac-address	ipv6	ipv4	ansible-role
server1	4c:72:b9:7c:41:17	2001:470:f325:503::5	10.128.3.5	core
server2	4c:72:b9:7c:40:ec	2001:470:f325:103::5	10.0.3.5	core
server3	4c:72:b9:7c:44:88	2001:470:f325:503::6	10.128.3.6	monitoring
server4	4c:72:b9:7c:43:27	2001:470:f325:503::7	10.128.3.7	automation
```

owendelong:
Roger all.

Thanks!

Owen

owendelong:
Thinking about this a little more… Would you mind configuring address per service?

Even though they’re currently going to be on the same machine, I’d like to set it up along the following lines:

	…3.A		Core 1 Machine Address — Used for management/Administration
	…3.B		DNS Server on Core 1
	…3.C		DHCP Server on Core 1
	…3.D		Log Server on Core 1
	…3.E		NTP Server on Core 1

I use A/B/C/D/E because I don’t care about the actual numbers at the end (or, for that matter, the order).

Just I’d like to see us have service addresses that can be moved to other machines independent of the machine
management address just in case we have some reason to do so.

If we’re too far down this road to do that, no biggie. But if it’s an easy thing to change, I’d appreciate it.
It just might save us a few headaches during the show. (Or it might be totally useless).

Owen

PR #28, opened 2018-02-17
[Ready] BIND in Ansible

kylerisse:
## Description of PR
Implementation of BIND with dynamic DNS zone file generation.

## Previous Behavior
No functional authoritative DNS in build

## New Behavior
BIND with scale.lan zone implemented. All host records (A, AAAA, and NS) created dynamically for servers and switches

## Tests
output of _inventory.py_ was reviewed as well as multiple rounds of _vagrant up_ and _vagrant provision_


kylerisse:
unbound is only designed to be a caching name server. in order to get it to support the scale.lan it gets quite hacky and wouldn't be ideal to support. I switched to bind as the first alternative because it is well known to the team. the main work was making sure i could feed the switch list in and interact with it to create a proper zone file. nsd is an alternative to bind that we could probably do a drop in later, with minimal tweaks to the jinja templates. 

tl;dr: it's more about getting the inventory.py and templates working than final decisions on the first pass of services, but unbound wont be sufficient which was my oversight.

sarcasticadmin:
@kylerisse ahh right on, good call!

PR #29, opened 2018-02-21
Owen v0.0.5 -- Review ONLY -- DO NOT MERGE

owendelong:
THIS IS A REVIEW REQUEST only. Do not merge yet. (I'm still actively working on the code.

Rob/Kyle, please review this and provide any feedback. Let me know if you find the code readable or feel that there are improvements needed so that someone besides me can understand  what is going on.

I've tried to modularize it rationally as I go, but I've also been rushing it to try and get it cranked out quickly, so there may be tradeoffs I didn't notice.

## Description of PR
Significant progress towards configuration file automatic generation.

## Previous Behavior
Partially generated configuration file interfaces section.

## New Behavior
Now fully generates interfaces and Layer 3 interface. Many bugs squashed.

## Tests
I ran the script and looked at the output.


kylerisse:
It's clear what you're attending to accomplish, I would feel pretty confident helping support it. I don't have any major changes to suggest. 

Everything below this is either a tangent or attempt at humor: 

I noticed you are undecided how we should handle privilege separation, as am I. I'd like to leverage the same auth keys for the linux servers as well. Perhaps it's a matter of having two different key folders, 1 for admins, and 1 for non-admins. like /authentication/keys/{admin,standard} ? In theory at least 1 admin, if not 2, will be approving any PR which would include keys. But then again, you're fond of pointing out the flaw of "in theory" vs "in practice", haha. 

oh and my sworn oath to pretending to be a hipster requires me to poo poo this commit as is commits 2 sins:
1. use of perl
2. use of ipv6

j/k, looks good Owen!

owendelong:
@kylerisse Thanks. I like the idea of {admin,operator} directories and sorting keys accordingly.

The linux privseparation is probably more subtle than that required for the routers/switches.

In the router/switch world, there's essentially super-user (read,write) and operator (readonly).

There are more granular and other privileges available, but the implementation overhead (RADIUS, etc.) and other issues lead me to believe it isn't particularly useful in our environment. I'm not even sure who we would put in the "operator" category now.

PERL is a virtue, not a sin, as is IPv6, so I'm not so worried about that comment. 👅 

Robert, any thoughts?

switch-configuration/config/scripts/template.pl
```diff
@@ -1,15 +1,139 @@
 #!/usr/bin/perl
```
sarcasticadmin: What do you think about `#!/usr/bin/env perl` so it compatible with the BSD folks?

owendelong: I think that would break Mac and most Linux installations.

Yes, the named references are intended. That's the usual Juniper way. The script doesn't yet write out the "vlans" section of the configuration file (I'm actively working on that code as I write this), which provides the association between numbers and names. (It also does the definitions necessary for private vlans, including community and isolated secondary private vlans.

owendelong:
I'm closing the pull request now that I have the requested feedback. I'll open a new pull request after I commit the version which writes the complete config file for a single test switch. Then I'll begin developing the code that will use this library (without the test generator) to iterate through and write out all of the switch configurations.

switch-configuration/config/scripts/template.pl
sarcasticadmin: @owendelong I dont think that would break anything. I can double check on my mac tomorrow if you want. I usually use `#!/usr/bin/env <interpreter>` for most things I write.

Thanks for the clarification about the the refs.

owendelong: OK... Tested it and it does appear to work on Mac and Fedora. Will make the change.

sarcasticadmin: hehe yay, thanks! My FreeBSD laptop appreciates it!

PR #30, opened 2018-02-21
[READY] Building LEDE from src

sarcasticadmin:
## Description of PR
**This is not ready to be merged yet. I still would like to create some automation and docs around the build process.**

This is the initial build of lede from src, specifically commit: `7da64807002cbe665176b125d8b2ee1a46d4e785`

Pkg changes are as follows:
```diff
- dropbear (default ssh daemon)
- logd
- dnsmasq
- firewall
+ openssh
+ openssh-client
+ rsyslog
+ zabbix
+ iperf3
+ serial drivers
+ lldp
+ snmpd
```

Will be adding more shortly just doing incremental validation since its hard to tell what breaks when building this all at once.

## Previous Behavior
Image builder of current stable lede version which is pretty old

## New Behavior
Latest and greatest

## Tests
Loaded onto 3700v2 and comes up just fine. Still working through the changes though.


sarcasticadmin:
@owendelong @davidelang @kylerisse this is ready to go.

to build an image all you should need to do is:
```
# Update secrets/openwrt.yaml
make templates
make build-img
```

owendelong:
Perhaps this should go in a BUILD.md or README.md file in the tree?

Owen

sarcasticadmin:
@owendelong good point, its in the [README.md](https://github.com/socallinuxexpo/scale-network/blob/master/openwrt/README.md#image-with-templates) But its outdated with the old build instructions. So Ill update that on my next PR.

PR #39, opened 2018-02-27
[READY] Openwrt template network config

sarcasticadmin:
## Description of PR
We can now build network configs using `gomplate` so that they are included into the openwrt img build. @owendelong I was able to successfully bridge the vlans so that we dont need to have separate images or startup scripts to switch between the vlan differences in the expo vs conference halls.

## Previous Behavior
* network config was not included in image outside of defaults.
* `config/` was being put into `/`

## New Behavior
* network config now matches vlans required for conference and expo.
* network config also templated out in `gomplate`
* `config/` now correctly added to `/etc/`

## Tests
1. Added vlan `103` for mgmt with dhcp to my local machine and was able to successfully ssh
2. Added vlan `108` for staffwifi with dhcp to my local machine. Was able to connect to wifi and use the network through AP.
3. Validated `vlan` config while ssh'd into AP: `swconfig dev switch0 show`
4. Validated `wifi` radios while ssh'd into AP: `wifi status`


owendelong:
Can you be more specific about “bridge the VLANs”?

I want to make sure we don’t run into a situation where we lose our layer 3 separation in an unexpected way.

Owen

sarcasticadmin:
@owendelong sure, essentially this was what we discussed on one of the networking calls with @davidelang. Bridging multiple vlans on an interface was @davidelang's proposed idea and the one I thought was most practical to pursue due to the layer 3 separation. So what I was able to do was create the interfaces in openwrt to include both vlans pairs. For example: `2.4G Wireless Network in Expo Center` interface will have vlans `100` & `500`(https://github.com/socallinuxexpo/scale-network/blob/master/facts/secrets/openwrt.yaml.example#L25) That way if either vlan exists on the switch we'll get connectivity. This shouldnt be a problem since those vlans will never exist on the same switch due to the layer 3 separation.

owendelong:
OK… Got it.  Thanks.

Owen

Issue #15, opened 2018-02-11
Create Mechanism for turning off AP radios in LEDE


sarcasticadmin:
Create the mechanism to turn off the AP radios if it cannot reach the default gateway. Probably easiest to build a shell script thats periodically called via cron or from a service.

PR #42, opened 2018-03-01
[READY] wifi gateway monitor and then some

sarcasticadmin:
## Description of PR
This allows us to monitor gateway endpoints and turn off wifi networks when the gateways are not reachable. Fulfills #15 

## Previous Behavior
Wifi radios were always advertising the wireless networks even if the network was unhealthy

## New Behavior
* Wifi networks are taken down when the endpoint is not reachable
* Added @davidelang's wifi report script
* Added logrotate package and config
* Added rrdtool1 package
* Resolver is now configurable
* Fixed some `Makefile` issues

## Tests
Tested on `3700v2` and this will be the image used for testing on sunday

### DIFF
--- a/facts/secrets/openwrt.yaml.example
+++ b/facts/secrets/openwrt.yaml.example
@@ -8,6 +8,11 @@ rsyslog:
 zabbix:
   server: 'server3.scale.lan'
 
+nameserver: '8.8.8.8'
+
+apinger:
+  endpoint: '8.8.8.8'
+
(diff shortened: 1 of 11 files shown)


sarcasticadmin:
@owendelong @kylerisse this is ready to go!

kylerisse:
Awesome, thanks @sarcasticadmin 

enjoy the wild weekend!

PR #47, opened 2018-03-02
Owen v0.1.1

owendelong:
## Description of PR
Adds descriptions for support for routers to README file

## Previous Behavior
No router support in README

## New Behavior
At least we're talking about it now.

## Tests
Yak Yeti Yak Don't come Back.


sarcasticadmin:
👍 thanks for adding some bits about the routers!

PR #56, opened 2018-03-05
Update switch_template.pl

owendelong:
Added management ethernet down alarm bypass.

## Description of PR
Puerto Rico is an Island possession of the United States of America in the Caribbean.

## Previous Behavior
They've repeatedly requested statehood with violent protests.

## New Behavior
They're trying to get the electricity back on after a hurricane.

## Tests
We applied generous amounts of wind and water until the infrastructure failed.


sarcasticadmin:
⏰

PR #124, opened 2018-12-17
[WAIT] Maps of setup

sbibayoff:
## Description of PR
Started a Maps directory, include last years setup maps.

## Previous Behavior
no maps located

## New Behavior
getting maps into tech github so everyone could see

## Tests
How was this PR tested?
Going to see if it makes it in


sarcasticadmin:
@sbibayoff thanks for round these up! Can we see about putting this in the `github wiki`, `googledocs`, or `s3` instead? Just because these are binaries and I would prefer to have them outside of `git`. We can still make references to them within a `readme.md` similar to the juniper images: https://github.com/socallinuxexpo/scale-network/tree/master/switch-configuration#ex4200

what do you think?

sarcasticadmin:
@sbibayoff merged #126 for now but still hoping we can get access to the gdrive in the future. Ill close this so theres no confusion

PR #126, opened 2019-01-15
[READY] Adding links to maps pdfs

sarcasticadmin:
## Description of PR
@sbibayoff I added these to `s3` instead of directly in git. That way theres no having to mess with `git-lfs`. This was my alternative to #124. Let me know what you think.

Sorry for the delay.

## Previous Behavior
We didnt have maps

## New Behavior
Now we have maps for the Pasadena Convention Center from 16x 

## Tests
Links work

### DIFF
--- a/MAPS.md
+++ b/MAPS.md
@@ -1,0 +1,15 @@
+# Maps
+
+Maps of the layout for SCaLE 16x at Pasadena Convention Center maps can be found:
+
+* [cbll-diagram-1](http://sarcasticadmin.com/scale/maps/16x/cbll-diagram-1-scale.pdf)
+* [cbul-diagram-2](http://sarcasticadmin.com/scale/maps/16x/cbul-diagram-2-scale.pdf)
+* [cbul-diagram-3](http://sarcasticadmin.com/scale/maps/16x/cbul-diagram-3-scale.pdf)
+* [ballroom-diagram-4](http://sarcasticadmin.com/scale/maps/16x/ballroom-diagram-4-scale.pdf)
+* [ballroom-diagram-5](http://sarcasticadmin.com/scale/maps/16x/ballroom-diagram-5-scale.pdf)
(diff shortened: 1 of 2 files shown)


sbibayoff:
Hey Rob,

This works. But would it be better to have them on the wiki as one of the suggestion you made? I don't know if scale has an s3 account, but would prefer to have all scale stuff under a scale "property".

just a thought

sarcasticadmin:
@sbibayoff I totally agree. I just put the pdfs in my public s3 bucket for the time being since that's also where the firmware for the switches currently resides: https://github.com/socallinuxexpo/scale-network/tree/master/switch-configuration#firmware

I raised the issue on the tech mailing list, we'll see what we can do about getting it into some that Scale itself can manage.

sarcasticadmin:
@owendelong we actually dont want to `merge` master into `feature` branches. Typically if we need to Id recommend `rebasing`. `git` doesnt really make this obvious though. Ill take care of undoing your `merge` on this branch and then we can merge this for now

sarcasticadmin:
@owendelong I didnt realize there was a merge conflict with the `README.md` and thats why you merged `master`. Its a really annoying feature of github since it convolutes the history. Ill show you what I mean during the show. Regardless this is all fixed and Ill `merge`

owendelong:
Ah… OK… Now it makes sense. Sorry if I created a problem. Was confused by it, but decided to try and do something useful rather than just leave it sitting around untouched.

Thanks for cleaning up my gitmess again. ;-)

BTW, how would you feel about pushing to migrate to bitbucket instead of GitHub?

The reason I suggest this is that bitbucket has IPv6 support while GitHub doesn’t.

Owen

sarcasticadmin:
No worries @owendelong! I appreciate the effort and its something I really wish was clarified. If I cant read the `git log` then whats the point! `rebase` makes this possible

Umm I havent had the best experience with bitbucket, but if you want to raise this on the tech mailing list to get the discussion going Ill certainly give my .02 cents

PR #130, opened 2019-02-09
[READY] Circle CI start

kylerisse:
## Description of PR
Adds Circle CI integration to this repo

## Previous Behavior
No Circle CI integration

## New Behavior
- Lints ansible playbooks and python dynamic inventory script
- Runs a test build of openwrt

## Tests
Pushing to this branch and checking the Circle CI webui


sarcasticadmin:
@kylerisse Do you think we can also include the build badge in the README.md? I really like all the work on this circleci config!

kylerisse:
Yea great idea. I’ll get that going and get some docs added. This PR is a placeholder for our circle work so feel free to merge into it as well and make drastic changes if needed. Not married to anything just playing with it. The openwrt stuff I have is a placeholder and could definitely use some feedback. I omitted caching and attempting to run parallel builds for now. Circle is kicked off for every push repo wide but once we get work done here we can decide as a team how to filter by branch name or w/e.

kylerisse:
apologies for the horrendous commit count, unfortunately need to push to test changes. the `test.sh` for gomplate keeps failing with exit 1 due to missing `golden/root/.ssh/authorized_keys`. stdout is blank so it's hard to catch the problem. verified same condition locally using same process in docker. 

Might consider:
1. modifying the diff in test.sh to `diff -N -r golden/ tmp/`
2. modify `-u` to include hidden files
3. rerun `test.sh -u` to capture the missing files in the current golden and commit

@sarcasticadmin I'd like you to double check my findings rather than me resolving using these steps in case you want something different

sarcasticadmin:
@kylerisse wow thanks for getting this so far! I totally understand the plethora of commits to get this going.

I could have sworn I was including `-u` in `diff` as well so it was easier to see the difference between `golden` and `tmp`. But now that I think about it more, `-u` might have been being picked up from my local `.bashrc`. Ill look into the hidden files too since it weird that I wasnt seeing that locally either. local testing fails :-1: Regardless I think your suggestions are going to be the case to fix these failures

sarcasticadmin:
@kylerisse Fixed the gomplate tests! Looks like it was a combination of your recommendations (.ssh file was being ignore due to .gitignore) as well as leverage the `gnu diff` that comes from alpine's `diffutil` pkg.

I went ahead and reworked the commits as well.

Let me know what you think!

kylerisse:
@sarcasticadmin This is awesome! Thanks for cleaning the git mess up. And the PR is passing its own tests too.

Issue #144, opened 2019-03-01
Make sure Changes from branch od.17.01 don't break Ansible integration


owendelong:
Kyle,

In checking in od.17.01 (or was it OD.17.01?), I changed the content of some of the tsv files. I don't think I did anything that will break ansible and you can look at the diffs for the README file that describes them to see what changed (I did update all the documentation too).

Please make sure, however, that I didn't break any of your scripts with my changes and/or update your scripts accordingly.

kylerisse:
I did check those and looks good. The columns in switch types are the same, just with an added column. Ansible wasn't using anything in the booth_list so we're good there. Also verified the json being spit out looks sane. https://circleci.com/gh/socallinuxexpo/scale-network/90 

The DHCP for the vendors was handled by the router last time. I wont need to do anything if we follow that model again.

owendelong:
> On Feb 28, 2019, at 20:19 , kylerisse <notifications@github.com> wrote:
> 
> I did check those and looks good. The columns in switch types are the same, just with an added column. Ansible wasn't using anything in the booth_list so we're good there. Also verified the json being spit out looks sane. https://circleci.com/gh/socallinuxexpo/scale-network/90 <https://circleci.com/gh/socallinuxexpo/scale-network/90>
> The DHCP for the vendors was handled by the router last time. I wont need to do anything if we follow that model again.
> 
> 
Yeah, we didn’t follow that model this time.

You’ll need to generate DHCP for the Vendor VLANs built by the combination of VVRNG and VVLAN directives.

It’s a bit of a complex process (sorry) and I’m happy to talk it over with you on the hangout if you have questions.

Owen

kylerisse:
Seems strait forward enough. Is there already a working implementation of this in the perl scripts? When I run build_switch_configs.pl from master I don't see any vendor VLANs defined in Expo switches. My current assumption is that I will not until booth_list.tsv is populated and that functionality is added. Is that something #150 addresses?

owendelong:
Yes.. There is… I find it interesting that you don’t see them… When I run it, there definitely is.

If you look at output/Expo1A.conf, I see the following matching lines:

kiev:owen (360) ~/SCALE/networking/scale17x/scale-network/switch-configuration/config % grep -n -i vendor output/ExpoA1.conf                         2019/03/01 11:39:24
[...]

PR #148, opened 2019-03-01
[READY] Fix DHCP issue caused by Expo file changes

kylerisse:
## Description of PR
During further investigation for #144 found an error with DHCP services due to changes in the Expo vlans.d file. This fix eliminates the cause by testing if the vlan ID is a digit. Another solution would be to comment out the offending line in that file but this PR can proceed regardless of that choice.

## Previous Behavior
isc-dhcp-server would fail on start due to an invald entry in `/etc/dhcp/dhcpd.conf.subnets` related to the ex_v_ line in `vlans.d/Expo`

## New Behavior

- Inserted a quick check when parsing the vlans.d files to test whether the vlan ID is a digit. There isn't a valid case where a real VLAN would be a non digit (in the case ofo ex_v_ containing a "-") so there's no harm in adding it.
- Reduced the number of characters in line 77 to pass pylint, unrelated to bug fix.

## Tests
1. Ran `inventory.py` and verified the culprit VLAN was no longer present in the output.
2. After a `vagrant provision` verified `/var/log/dhcp.log` did not show exiting for v4 anymore.
3. Verified service health with `systemctl status isc-dhcp-server`. Service no longer in `failed` state.

### DIFF
--- a/ansible/inventory.py
+++ b/ansible/inventory.py
@@ -36,9 +36,12 @@ def populatevlans():
                 ipv4bitmask = ipv4[1]
                 ipv4dhcp = dhcp4ranges(ipv4prefix, int(ipv4bitmask))
                 ipv4netmask = bitmasktonetmask(int(ipv4bitmask))
+                vlanid = elems[2]
+                if not vlanid.isdigit():
+                    continue
                 vlans.append({
                     "name": elems[1],
-                    "id": elems[2],
+                    "id": vlanid,
                     "ipv6prefix": ipv6prefix,
                     "ipv6bitmask": ipv6bitmask,
                     "ipv4prefix": ipv4prefix,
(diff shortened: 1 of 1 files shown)


owendelong:
Hi, Kyle,

You shouldn’t be treating that line as a VLAN managed by ISC DHCP.

It’s a new directive (VVRNG) that defines a range of VLANs for Vendors.

These VLANs are local to a particular switch and the routing and DHCP for them is handled directly on the switch in question.

(See the VVLAN directive in the types/Booth file for how this range is used).

Concerns for effects like this is one of the reasons I created issue #144 and assigned it to you.

Generally, however, your DHCP parser should specifically be looking for VLAN directives and probably ignoring any other directives unless it specifically
knows what to d with them. I’ll take a look at your changes and provide additional feedback if I have any.

Owen

owendelong:
Upon reviewing the change, I see that this will work for now and so a proper solution isn't urgent, but I am going to file an issue requesting that we parse the directives in the file appropriately and not simply make assumptions that every directive is the same.

PR #156, opened 2019-03-02
[READY] Perl linting and perl ci

sarcasticadmin:
## Description of PR
Perl linting and testing

Add `-M'warnings FATAL=>q(all)'` in the perl compile and warning test to make sure that any warnings actually are treated as errors (non-zero return). Also leveraged `xargs` to make sure that we dont swallow the error like `find -exec`

## Previous Behavior
Didnt exist

## New Behavior
Adding linting and running the actual script to make sure it works

## Tests
How was this PR tested? CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -85,9 +85,35 @@ jobs:
         name: test inventory.py
         command: ./inventory.py | jq
 
+  perl_lint:
+    docker:
+    - image: perl:5
+    resource_class: medium
+    working_directory: ~/scale-network
+    steps:
+    - checkout:
+        path: ~/scale-network
+
+    - run:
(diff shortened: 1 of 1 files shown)


kylerisse:
link #143

sarcasticadmin:
@kylerisse i didnt mean to snake this one from under you, I was just playing around with the circleci config and then kinda got it all working around the same time you did lol

Anyway if your good with this, then so am I. Moving to `[READY`]

kylerisse:
@sarcasticadmin no worries at all!!! Was working through issues list. So long as things work doesn't matter how it gets there, imo. Must be a similar vibe flowing up and down the coast at that time, haha.

Issue #149, opened 2019-03-01
ansible/inventory.py ignores directives and makes invalid assumptions


owendelong:
In PR #148, the change brings to light the fact that the VLANs files are not being parsed correctly by ansible/inventory.py.

The intent is to parse the "vlans" file only. If that file contains "#include" directives, then only files identified by those directives should be parsed. The switch configuration generator will not process files from "vlans.d" which are not referenced in such a manner and neither should inventory.py.

Further, the directives in VLAN configuration files (whether the master "vlans" file or one of the included files from "vlans.d/" (currently) fall into two categories with different syntax.

Currently, inventory.py simply assumes that every directive is "VLAN" which is invalid. The syntax for a "VVRNG" specification is different and serves a different purpose. This line should not be commented out as it affects the config generation for switches affected by a file in "types/" which contains a "VVLAN" directive (currently only "types/Booth").

The correct approach is to compare the first field of each line (directive) in the file to a known directive understood by the parser. If the directive in question is not understood, the safest action for the parser to take is to emit a warning and ignore the line.

owendelong:
See https://github.com/socallinuxexpo/scale-network/blob/master/switch-configuration/README.md for complete documentation of the configuration files and how they should be parsed and interpreted as well as documentation of all valid directives.

kylerisse:
Good call, we should process the file as intended. This should come after #144 and #153 .

kylerisse:
Suggestions for added error handling https://github.com/socallinuxexpo/scale-network/pull/158#discussion_r261841739

owendelong:
I believe this is resolved... Kyle, can you verify and if so, close this issue?

PR #158, opened 2019-03-02
[READY] Handle VLAN or VVRNG directives

kylerisse:
## Description of PR
Modification to `inventory.py` to support parsing VVRNG and dynamically generating VLAN information for the range specified. Documented in #144 . The change isn't the greatest way to handle this long term but was done in such a way to limit changes across other parts of the script. It's basically a feature flag around the directive so all non-VVRNG code paths remain effectively unchanged and any errors can be found in a single place. An overhaul of this functionality should be tackled as part of #149 .

## Previous Behavior
Skipped VVRNG lines when parsing vlans.d files as that concept didn't exist until two days ago.

## New Behavior
Using the range and supernets specified in the VVRNG line simply calculate all possible values in the range and create our standard vlan dictionary for them and stick them in vlans dictionary so the jinja templates can consume them.

## Tests
1. Review the output of `inventory.py`, which can be done easily at https://circleci.com/gh/socallinuxexpo/scale-network/134. The big thing to check is that the ex_v_$VLAN vlans exist, the data looks sane, and boundaries have correct values. (ie vlan456 / 10.3.0.0/24, vlan200 / 10.2.0.0/24, vlan455 / 10.2.255.0/24, etc) Also verify the DHCP ranges make sense and match our standard dynamic splits (ipv[4,6]dhcp[1..2][a..b] and the subnet mask and gateway fields).
2. Spin up vagrant and do extensive review of `server2` and `server5`. Check `/etc/dhcp/dhcpd.conf.subnets` and verify all previous and new subnets look ok. Same with `/etc/dhcp/dhcpd6.conf.subnets`
3. On both servers verify function of `isc-dhcp-server` through systemd and the log at `/var/log/dhcp.log`. Please note that isc-dhcp-server6 does not work when testing locally as it has no interfaces to bind to on valid subnets.
4. On both servers verify sanity of `/etc/bind/named.conf.scale-acls` and the plethora of new Expo subnets.
5. Basic spot checking of DNS functionality across all boxes.
6. Idempotence testing with multiple `vagrant provisions` establishing stability of generated data

### DIFF
--- a/ansible/inventory.py
+++ b/ansible/inventory.py
@@ -5,6 +5,7 @@ SCaLE specific text files to produce a sane inventory to ansible
 '''
 import ipaddress
 import json
+import math
 import os
 import re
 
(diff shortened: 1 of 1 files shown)


ansible/inventory.py
```diff
@@ -27,42 +28,82 @@ def populatevlans():
...
-                    "ipv4dhcp2b": ipv4dhcp[3],
-                    "ipv4router": ipv4dhcp[4],
-                    "ipv4netmask": ipv4netmask,
-                    "ipv6dns1": "",
-                    "ipv6dns2": "",
-                    "ipv4dns1": "",
-                    "ipv4dns2": "",
-                })
+                directive = elems[0]
+                if directive == "VLAN":
```
sarcasticadmin: @kylerisse Should we have a fall through case? Im just wondering if an `else` would be a good idea if for some reason there was a `directive` other than "LVAN" or "VVRNG"

sarcasticadmin: Maybe even just an `else` that raises an error?

kylerisse: We're only concerned with these two directives here. There's nothing to do in this context should anything else show up. The suggestion from the last paragraph in https://github.com/socallinuxexpo/scale-network/issues/149#issue-416179334 is to ignore and emit an error. I'd prefer not to plumb in a separate error logging facility at this time to reduce the number of last minute changes. I could throw it to STDERR but the user will never see it anyways since ansible is redirecting it's output to itself, at least that's my assumption as I'd have to do further testing. It's also expecting a pure json blob which could be broken throwing extra strings into the mix. The downside isn't worth the risk for this week, imo. When tackling #149 that's something that can be explored though, logging facilities and more appropriate handling of directives.

kylerisse:
thanks @sarcasticadmin

ansible/inventory.py
sarcasticadmin: Ahh gotcha, good points!

owendelong:
I think on Kyle's side of the equation parsing these files, a warning would be the best case for an else statement. "Warning: Unknown directive (<directive>) encountered in file <filename>, skipping." or something like that. However, I don't think adding that should hold up this merge (and it obviously hasn't) ;-)

owendelong:
Excellent work, Kyle!

kylerisse:
You must. I don't need anything in `types`, only `vlans.d`. In those files the directive is either `VLAN` or `VVLAN`.

PR #166, opened 2019-03-05
Update README.md

owendelong:
More markdown syntax fun


sarcasticadmin:
@owendelong thanks for adding this!

@sbibayoff thanks for approving!

PR #195, opened 2019-03-07
[Ready] add steveb to server access

kylerisse:
## Description of PR
Add Steve B to server access

## Previous Behavior
Steve B's key had invalid name and no user accounts

## New Behavior
Steve's key and user account is now managed by Ansible

## Tests
Live in NOC


sarcasticadmin:
DO IT LIVE :shipit:

Issue #182, opened 2019-03-06
openwrt turn on WPS LED on startup


davidelang:
we need to be sure the build finished properly, so since the 17x image leaves the light off, make the 18x build turn it on (or blink something) so we know at the end of flashing that we have the new build instead of the old one

davidelang:
page that details how to configure LEDs in openwrt
https://openwrt.org/docs/guide-user/base-system/led_configuration

PR #262, opened 2019-12-04
[READY] issue-182 - Adding WPS LED build identifier

sarcasticadmin:
## Description of PR
Fixes: #182 

This allows us to be able to have a visual cue about whether or not the image was flashed correctly. 
 
## Previous Behavior
* WPS LED was always off

## New Behavior
* For even scale conferences the WPS led will be `on` and for odd scale conferences it'll be `off`
* Moved ntp config to `facts`
* Added tests for ntp config serverspec

## Tests
* Flashed and validated that WPS LEDs are ON as expected on both ipq806x and ar71xx builds
* Also tested the inverse for odd scale numbers where WPS LED is OFF

### DIFF
--- a/facts/secrets/ar71xx-openwrt-example.yaml
+++ b/facts/secrets/ar71xx-openwrt-example.yaml
@@ -14,6 +14,9 @@ zabbix:
 
 nameserver: '8.8.8.8'
 
+ntp:
+  server: '0.openwrt.pool.ntp.org'
+
 apinger:
   endpoint: '8.8.8.8'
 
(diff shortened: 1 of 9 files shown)


davidelang:
this would not require a new daemon, there is a morse code mode in the LED 
subsystem aht lets you set a message and it will cycle it.

owendelong:
Morse would potentially work as well. I know you and I are hams that are old enough to read Sam’s numbers, but how many other team members could?

Owen

davidelang:
I would just do something like 'last digit of the build #" what we're looking 
for is something easy to identify if the flash worked or not. we don't need a 
lot of info.

David Lang

davidelang:
but as you say, this is good enough for this year.

sarcasticadmin:
@davidelang @owendelong I like the idea, willing to discuss in more detail later

PR #265, opened 2020-01-06
[READY] auto-generate printed switch maps and router backups

owendelong:
## Description of PR
Fixes: #266 

Adds the ability to auto-generate printed switch maps (PostScript files) from configuration
Also includes backups of the router configuration files and some checkpointing of new scripts that aren't yet in use.

## Previous Behavior
This feature did not exist

## New Behavior
This feature works

## Tests
Manual review of a complete set of generated PS files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,9 @@ tests/unit/openwrt/tmp*
 # Ignore built configuration files directory
 switch-configuration/config/output
 
+# Ignore built switch map files
+switch-configuration/config/switch-maps
+
 # Created by https://www.gitignore.io/api/osx,vim,emacs,windows,libreoffice,sublimetext
 
 ### Emacs ###
--- a/router-configuration/backups/br-mdf-01
+++ b/router-configuration/backups/br-mdf-01
@@ -1,0 +1,383 @@
+## Last changed: 2019-10-27 02:59:16 UTC
+version 15.1X49-D70.3;
+system {
+    host-name br-mdf-01;
+    root-authentication {
+    }
+    name-server {
+        8.8.8.8;
+        8.8.4.4;
+    }
+    login {
+        user dlang {
+            uid 2002;
+            class super-user;
+            authentication {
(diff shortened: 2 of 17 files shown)


.gitignore
```diff
@@ -10,6 +10,9 @@ tests/unit/openwrt/tmp*
 # Ignore built configuration files directory
 switch-configuration/config/output
 
+# Ignore built switch map files
+switch-configuration/config/switch-maps
```
sarcasticadmin: Thanks for adding the .ignore

router_configuration_backups/br-mdf-01
```diff
@@ -0,0 +1,385 @@
+## Last changed: 2019-10-27 02:59:16 UTC
```
sarcasticadmin: can we move these to under `switch-configuration`? Something like `switch-configuration/config/backups`  since I feel liek that would make more sense then the toplevel dir here

router_configuration_backups/br-mdf-01
```diff
@@ -0,0 +1,385 @@
...
+version 15.1X49-D70.3;
+system {
+    host-name br-mdf-01;
+    root-authentication {
+        encrypted-password "$5$.2JOEO0i$GaaUUIlcgelNVJ7qhfu1jnISVcALe0CtWAoRdn.H1a9";
```
sarcasticadmin: Can we scrub these from the commit history? Even though we wont be using them Id rather just not include them to at all

router_configuration_backups/br-mdf-01
owendelong: Well. We can, but it goes against my network admin view that there’s a distinction between switches and routers. Everything under switch-configuration relates to auto-generation of config a for the ex-series switches. The router config are manually maintained at least for now. 

As such, having it separate at top level actually makes more sense to me.

router_configuration_backups/br-mdf-01
owendelong: A little confused. These are the historical configs from the 3 srx routers. I might need to do some cleanup if I have them in more than one place (oops) but I do want to keep records of them in the commit history.

sarcasticadmin: Sorry I was talking about the actual line here for `encrypted-password` I know well be rotating this for 18x but still Id rather just have it be junk vs something real

router_configuration_backups/br-mdf-01
sarcasticadmin: Fair enough then can it at least follow the existing naming convention? So `router-configuration/`

O and arent EX-series switches are L2/L3?! :smile:

router_configuration_backups/br-mdf-01
owendelong: Hey. Sure. I can do that later tonight.

owendelong: OK... Added a commit removing the passwords.

router_configuration_backups/br-mdf-01
sarcasticadmin: Any thoughts on the dir?

sarcasticadmin:
@owendelong i went ahead and `fixedup` the secrets commit so that the commit history wouldnt contain the hashed secret anymore.

I also moved the router backups to `router-configuration/backups` so the toplevel dir conforms with the existing `switch-configuration` dir.

Going to merge and thanks adding the ability to generate host `ps` files!

PR #272, opened 2020-01-21
[READY] Issues 149 and 258

### LINKED ISSUE #149: ansible/inventory.py ignores directives and makes invalid assumptions

### LINKED ISSUE #258: Make CSV files compatible with GitHub beautification
There's a standard for header lines and documentation lines that GitHub expects in CSVs.

If we make our CSV files compatible with those, then GitHub will beautify them for us when viewed in the web interface. This would be useful.

It's not urgent.

kylerisse:
## Description of PR
Fixes #149 
Fixes #258 

Fixes for inventory.py related issues.

## Previous Behavior
- inventory.py didn't parse the vlans source files as they were intended, didn't support the #include directive
- csv files were not git beautified
- pylint was set to a less strict mode due to inefficiencies in inventory.py

## New Behavior
- inventory.py now properly parses the vlans as intended and supports #include per spec
- csv files are git beautified
- pylint is now set to default strict mode

## Tests
- examining the inventory.py file output using jq, comparing to previous version output
- running pylint without the --confidence=HIGH flag to verify 10/10
- vagrant up; vagrant provision and making sure ansible still runs as expected
- ssh into the vagrant managed vms and spot check the dhcp and dns configurations plus logs
- reviewing 4 csv files in the github ui to verify git beautification works (listed below)

- https://github.com/kylerisse/scale-network/blob/issue149/facts/aps/aplist.csv
- https://github.com/kylerisse/scale-network/blob/issue149/facts/pi/pilist.csv
- https://github.com/kylerisse/scale-network/blob/issue149/facts/routers/routerlist.csv
- https://github.com/kylerisse/scale-network/blob/issue149/facts/servers/serverlist.csv


kylerisse:
> Going forward, what are your thoughts on unit tests for `inventory.py`? Im not expecting anything before Scale 18x but it seems like these functions are pretty particular and it'd be easy for someone else (me, mostly) to miss something if they were to try and help in this area of the codebase.
>
> On a slightly different topic, I also think it might be worth putting into place some simple `awk` tests to validate that the `.csv`s have the right number of delimiters.

I agree @sarcasticadmin . A more robust test harness around this stuff is a great idea and probably long overdue. Look forward to bouncing some ideas back and forth this conference season.

sarcasticadmin:
:shipit: thanks again @kylerisse for getting this done!

PR #284, opened 2020-02-21
[READY] Adding more templates and updating docs

sarcasticadmin:
## Description of PR

Adding more specific bhyve templates for the vms and docs that Ive corrected while working with the servers more

## Previous Behavior
* Docs were a little bit lacking
* Templates didnt have disk sizes

## New Behavior
* Docs for bhyve are better
* templates have disk sizes and a few more types
* updating servers.csv with new vm macs

## Tests
Validated on `bhyve1`


kylerisse:
@sarcasticadmin @owendelong Please role this PR back, it completely breaks ansible. Please use PR process to test.

Use vagrant, run vagrant up, when ansible kicks off you'll see the following from the serverlist changes. The first column is the servername and the last is the role. These should not be the same:

[WARNING]: Found both group and host with same name: core
[WARNING]: Found both group and host with same name: signs
[WARNING]: Found both group and host with same name: monitoring
[WARNING]: Found both group and host with same name: automation

sarcasticadmin:
@kylerisse Ill deal with this shortly, sorry about that. Now I see where it went a little sideways.

kylerisse:
@sarcasticadmin actually I see a problem with my testing vs the testing on bhyve. You can leave this. It makes sense that it applied for you but not for vagrant. I was incorrect it doesn't completely break Ansible which is why it works for you. It breaks it in vagrant because we haven't added bsd hosts to https://github.com/socallinuxexpo/scale-network/blob/master/ansible/Vagrantfile . Let's clean up tomorrow. This PR should be fine. The errors of group/hostname being the same is a red herring, we can discuss that tomorrow.

sarcasticadmin:
> How are we doing the IPv6 networking on these (networking in general)? Is it a bridged environment or are we creating routed subnet(s) for the virtual hosts?

The VMs are on a bridge with tap interfaces. The bridge is attached to `igb0` so the vms appear on the same subnet has the host (our management vlan). ipv6 is working as well. Sorry `/etc/rc.conf`already contains the ipv6 config from the base install which is why its not in the docs (but would be worth adding honestly).

> Do the hypervisors need IPv4? If so, for what?

I dont think they do. So we could definitely turn that off once we correct some of the ansible issues

PR #290, opened 2020-02-23
[READY] add new signs/reg image instructions for pi4

### LINKED ISSUE #256: Create build procedure for Pi 4B
Create and document build procedure for Pi 4B images

kylerisse:
## Description of PR
Fixes #256 

## Previous Behavior
Only had pi4 images for "imaging" and "general" pis, none for registration / signs.

## New Behavior
Have pi4 images for registration / signs. Instructions added.

## Tests
Building pis at work party. Having David and Peter try to do nefarious things. Creating compressed image.

### DIFF
--- a/facts/pi/gold_image.txt
+++ b/facts/pi/gold_image.txt
@@ -1,115 +1,93 @@
-# Steps used to build the gold image
+# Steps used to build the gold image for signs / reg
+
+Instructions heavily influenced by https://youtu.be/T9AtKld8USU
 
 # Create bootable SDCARD:
 1. Download Raspbian Stretch Lite 
-  a. 2017-11-29-raspbian-stretch-lite.img
-2. Write to 4GB SD Card using Etcher
+  a. 2019-09-26-raspbian-buster-lite.img
+2. Write to an SD Card using Etcher
 
-# Boot and update system:
+# Boot and setup system:
 1. Boot up
 2. Log in as pi/raspberry
-3. run `sudo apt-get update`
-4. run `sudo apt-get upgrade -y`
-5. run `sudo apt-get dist-upgrade -y`
-6. run `sudo apt-get install -y \
-    git \
-    openbox \
-    xwit \
-    python-gtk2 \
-    python-webkit \
-    xinit \
-    zabbix-agent \
-    lightdm \
-    nodm \
-    matchbox-window-manager
-7. run `sudo reboot`
-8. Log in again after reboot
+3. sudo adduser admin (set password)
+4. sudo usermod -aG sudo admin
+5. sudo apt-get update; sudo apt-get upgrade -y
+6. sudo reboot
+7. Log in as pi
+8. sudo apt-get install lightdm -y
+9. sudo raspi-config
+-- 2. Network Options
+---- N1. Hostname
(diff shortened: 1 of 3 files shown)


facts/pi/scripts/reg.py
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env python
```
sarcasticadmin: NIce! Glad we've got a reg pi on the rpi4!

facts/pi/scripts/web.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python
```
sarcasticadmin: Would have expected this in `go` 😉 But cool this looks great!

facts/pi/gold_image.txt
```diff
@@ -1,115 +1,93 @@
-# Steps used to build the gold image
+# Steps used to build the gold image for signs / reg
```
sarcasticadmin: Thanks for the new instructions!

sarcasticadmin:
> Line 6 of gold_image.txt still refers to Stretch.
> 
> Perhaps we should:
> 
> 1. Download current specified Raspbian Lite image.
>    a. As of 2020-02-23 current is 2019-09-26-raspbian-buster-lite.img
> 
> In step 4, perhaps we should also change Language from EN_GB to EN_US.

@owendelong I think some of these might be too late since the image currently exists. I think we should just pin out a follow up to include a new image (post-scale18x) the has the language change. Since anything at this point will probably have to be manual.

facts/pi/scripts/web.py
kylerisse: haha! if it were new maybe, but it's just a copy of the same script we use every year. They used to live in the scale-signs repo but seem more appropriate with the instructions.

kylerisse:
> > Line 6 of gold_image.txt still refers to Stretch.
> > Perhaps we should:
> > 
> > 1. Download current specified Raspbian Lite image.
> >    a. As of 2020-02-23 current is 2019-09-26-raspbian-buster-lite.img
> > 
> > In step 4, perhaps we should also change Language from EN_GB to EN_US.
> 
> @owendelong I think some of these might be too late since the image currently exists. I think we should just pin out a follow up to include a new image (post-scale18x) the has the language change. Since anything at this point will probably have to be manual.

SSH is enabled on the signs and reg stations. We can push out changes over ssh / ansible at run time.

kylerisse:
> This looks good @kylerisse ! thanks for put it together and working this out over the week, definitely save us the pressure of trying to do this the week of the show.
> 
> I spun out #299 since going forward we'll be putting more than just data and inputs into the `facts/pi` dir. I think we should at least start to talk about making a `pi` dir at the top-level all of this existing work and stuff to come in the future. Let me know what you think, but we'll do this in a follow up post-18x

That's a good idea. I imagine whoever takes over maintenance after this year could look into something like buildroot or another software based repeatable solution. These instructions can be used as a base or scraped during that effort.

tweak42:
I tested out the build instructions on my Pi 3 and 4 and all the plymouth stuff from lines 42 - 53 does not work because the splash graphic needs to be a .png file.  Is a momentary boot splash screen for a kiosk really necessary?

davidelang:
no, we would prefer no splash screen.

David Lang

PR #296, opened 2020-02-24
[READY] - PostScript switch maps

owendelong:
## Description of PR
Finalizes switch-map generation process
Fixes an error in the number of ports configured on the cfAV switch
Adds a README.md to the scripts directory explaining each script
Fixes a typo in an error message in the configuration generator script

## Previous Behavior
Switch-maps were individual PS files ready for printing, one map per file and per page.

## New Behavior
Switch-maps are now individual EPS files. The generate_ps_maps.pl script will combine them into a single PS file which prints 5 maps per page.

## Tests
Tested with both PS2PDF and by printing the results to my printer.


switch-configuration/config/scripts/README.md
```diff
@@ -0,0 +1,29 @@
+# Switch Configuration Scripts
```
sarcasticadmin: I appreciate the docs! I think this will provide a much better understanding to those who might not be familiar with the switch-config scripts

PR #298, opened 2020-02-24
[READY] Add Jack List

owendelong:
## Description of PR
Store Wall Jack list obtained from SCaLE17x in Repo

## Tests
N/A documentation only

### DIFF
--- a/switch-configuration/config/WallJackList
+++ b/switch-configuration/config/WallJackList
@@ -1,14 +1,38 @@
 Jack	Device			Location
-NE8		BallroomC		South East Corner
+NE8	BallroomC		South East Corner
 NE26	BallroomB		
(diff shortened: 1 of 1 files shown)


sarcasticadmin:
@owendelong im not sure why there is a commit from David in this PR, but I think the is from an old branch on master as well. It looks like the tests its trying to run are old which tipped me off that this might be from an old copy of `master`. I think what I would do would be to update your local `master` and `rebase` this then push it back up:

```
git checkout master
git pull --rebase origin master
git checkout od.17.jacklist
git rebase master
git push origin od.17.jacklist -f
```

sarcasticadmin:
@owendelong confirmed this is good, thanks for adding this after you were pruning branches!

PR #300, opened 2020-02-25
Fix reference to Rapsbian version.

owendelong:
## Description of PR
Clarifies Raspbian version and eliminate reference to older version of Raspbian.
Adds mechanism for specifying future versions without losing track of previous specified versions.

## Tests
Documentation only -- N/A

### DIFF
--- a/facts/pi/gold_image.txt
+++ b/facts/pi/gold_image.txt
@@ -3,8 +3,8 @@
 Instructions heavily influenced by https://youtu.be/T9AtKld8USU
 
 # Create bootable SDCARD:
-1. Download Raspbian Stretch Lite 
-  a. 2019-09-26-raspbian-buster-lite.img
+1. Download the current specified Raspbian image from https://www.raspberrypi.org/downloads/raspbian/
+   a. As of 2020-02-24 the specified image is 2020-02-13-raspbian-buster-lite.img
 2. Write to an SD Card using Etcher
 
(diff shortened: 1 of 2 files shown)


facts/pi/gold_image.txt
```diff
@@ -3,8 +3,8 @@
 Instructions heavily influenced by https://youtu.be/T9AtKld8USU
 
 # Create bootable SDCARD:
-1. Download Raspbian Stretch Lite 
-  a. 2019-09-26-raspbian-buster-lite.img
+1. Download the current specified Raspbian image from https://www.raspberrypi.org/downloads/raspbian/
+   a. As of 2020-02-24 the specified image is 2020-02-13-raspbian-buster-lite.img
```
sarcasticadmin: @owendelong The change to line 6 makes sense to correct the `stretch` reference. However, which image did we actually use as our base on the pis for this year? I think that needs to be reflect for `a.` vs just the latest img as of writing this doc. So i believe that `a.` should still be: `2019-09-26-raspbian-buster-lite.img`

owendelong:
Agreed… My bad.

Owen

sarcasticadmin:
Thats ok, do you do want to correct the changes and then Ill merge?

owendelong:
OK… Done. I also corrected the extension on the filename from .txt to .md since the file is actually written in MarkDown.

Owen

PR #363, opened 2020-03-19
[WIP] Add PCC CAD files

irabinovitch:
## Description of PR

Add PCC CAD files.


davidelang:
please add one of these two since they include both buildings: 
Scale_14x_Cad_Rev_3_David_Lang.dwg
Scale_14x_Cad_Rev_3_David_Lang-orig.dwg

probbly the first, the second fails opening by some cd programs (as does the 
file that starts with #), I don't know why or if they include any additional 
detail.

David Lang

sarcasticadmin:
Thanks @irabinovitch for passing these along!

I think for the time being lets hold off on merging this into the repo since Id like to have it at the very least added via `git-lfs`, but I'm thinking ultimately well probably have these files (or which ever @davidelang is specifically looking for) stored in an object store along with the other artifacts once we get around to #286

sarcasticadmin:
Just so theres no confusion about different versions. These are the sha256sums for each of the CAD files:

```
$ shasum -a 256 *.dwg
bf4566e553a14e35dc6fcb3b876dbd898fc44e183c3039c0c9486a9d14df43f4  Zpasadena-cc-exp_GES.dwg
fd6f85989ce7a87d6f0027feff672218052b2b4f859c7ec62463edf9f11b21bc  Zpasadena-cc_GES.dwg
fd6f85989ce7a87d6f0027feff672218052b2b4f859c7ec62463edf9f11b21bc  pasadena-cc-lower and upper level_GES.dwg
```

Will be closing in favor of publishing to s3 https://github.com/socallinuxexpo/scale-network/pull/372

PR #365, opened 2020-03-24
[READY] Autoflash on actual WNDR hardware

sarcasticadmin:
## Description of PR
Putting up some of the work that was done during 18x around `autoflashing` the wndr3800ch.

There are a lot of things going on in the MR so Ill try to cover all of the methods to my madness:

## Previous Behavior
* Had to do flashing manually
* Configs for wndr builds didnt reflect conference configs

## New Behavior

1. First and foremost we have successfully integrated automated flash of the WNDR 3800ch that we modified during Scale 18x. It is being done via a gitlab-ci runner running on a raspberry pi 2 that we had laying around. Thats controlling a relay to toggle the power and the GPIO pins to "press" the reset button. This is all coordinated via `openwrt/autoflash` script being called via the gitlab ci pipeline.

2. The `gitlab-ci` pipeline is broken down into two components: `openwrt-ar71xx-build` and `openwrt-3800ch-integ`. For now `openwrt-ar71xx-build` should be ignored since the shared-runners provided by gitlab lack the oomph to build openwrt in a timely manner. It took about 2 hr 45min: https://gitlab.com/socallinuxexpo/scale-network/pipelines/139993816 . The logs from the job are also limited to 4mb: https://gitlab.com/gitlab-com/support-forum/issues/2790 so I ended up piping the output to `tee | grep` and just checking for errors. The build artifacts will include a complete `build.log` instead.

3. Going on the more relevant pieces of this PR: `openwrt-3800ch-integ` is the gitlab-ci job that needs to be triggered with the following environment variables: `OPENWRT_INTEG` set to `YES` and `WORMHOLE_CODE` set to a magic wormhole code (this will be optional soon). The idea here is to only trigger this job when one is ready and it allows the ability for the user to pass in a img of there choosing to `openwrt-3800ch-integ` which has always been the most awkward part of this process.

So what you can now do is something like pre-triggering the pipeline:

```
wormhole send openwrt-ath79-generic-netgear_wndr3800ch-squashfs-factory.img
On the other computer, please run: wormhole receive (or wormhole-william recv)
Wormhole code is: 6-bifocals-endow
```

Then trigger the pipeline in the UI setting the following setting:

```
Branch: Set your branch to the one that matches the PR
Environment variables: `OPENWRT_INTEG` = `YES` and `WORMHOLE_CODE` = `6-bifocals-endow`
Select run pipeline
```
Then the job will fire running `openwrt-3800ch-integ` on the pi and pull the image via magic wormhole.

Eventually I will add logic to `gather-artifact.sh` so that if `WORMHOLE_CODE` is not defined then it just uses the last image build on the current branch or lastest build off `master`. This will require us to reconcile where the builds are generated from (circleci, gitlab, etc.).

3. I am introducing a github actions workflow bot Im calling `tux` (open to suggestions here on the name). I thought would be useful to help trigger some of these builds like  `openwrt-3800ch-integ` in gitlab and do other tasks related to PRs in the future. This seems to be the easiest way to act on comment in a PR and that trigger other APIs. I dont think this gets registered until it lands in `master`: https://github.community/t5/GitHub-Actions/Workflow-files-only-picked-up-from-master/td-p/30348 but the idea would be in a PR you can comment: `tux openwrt-integ` which would trigger the workflow for the autoflash discussing in Item 2 against the existing branch. I testing the POST via curl directly, see below.

4. Reconcile the wndr default config to reflect what we have used in past Scale conference deployments. This is exactly what we used during scale-18x minus the root password and wifi passwords.

## Tests
* Successfully passed existing CI tests
* Gitlab private runner job: `openwrt-3800ch-integ` works as expected and includes `serverspec` tests: https://gitlab.com/socallinuxexpo/scale-network/pipelines/140025531
* Gitlab shared runner job: https://gitlab.com/socallinuxexpo/scale-network/pipelines/139993816 works but not going to be used until solutions to speed up are figured out. 
* Github actions test: `curl --request POST --form token=<redacted> --form "ref=sa/autoflash" --form "variables[OPENWRT_INTEG]=YES" --form "vari
ables[WORMHOLE_CODE]=7-detergent-pluto"  https://gitlab.com/api/v4/projects/17362342/trigger/pipeline` triggers the pipeline and works as expected. This is a close as I can test until the actions lands in `master`. See CI results: https://gitlab.com/socallinuxexpo/scale-network/-/jobs/527325894

### DIFF
--- a/.github/workflows/trigger.yml
+++ b/.github/workflows/trigger.yml
@@ -1,0 +1,34 @@
+name: PR comments to trigger gitlab-ci jobs
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  build:
+  # if comment string starts with "build"
+  # event must be a pull request (not a regular issue)
+    if: >
+      startsWith(github.event.comment.body, 'tux')
+      && startsWith(github.event.issue.pull_request.url, 'https://')
+      && github.actor == "sarcasticadmin"
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: 'Call gitlab API with event data'
+      id: event-data
+      run: |
+        set -eu
+        JOB=$( cut -d ' ' -f 2 <<< "${{ github.event.comment.body }}" )
+        JOB_ARG=$( cut -d ' ' -f 3 <<< "${{ github.event.comment.body }}" )
+        RESP=$(curl -sSf \
+          --url ${{ github.event.issue.pull_request.url }} \
+          --header 'content-type: application/json')
+        BRANCH=$(python3 -c "import sys, json; print(json.load(sys.stdin)['head']['ref'])" <<< "$RESP")
+        # Call gitlab
+        curl --request POST \
+          --form token=${{ secrets.GITLAB_TOKEN }}  \
+          --form "ref=$BRANCH" \
+          --form "variables[$JOB]=YES" \
+          --form "variables[WORMHOLE_CODE]=$JOB_ARG" \
+          https://gitlab.com/api/v4/projects/17362342/trigger/pipeline
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,45 @@
+---
+#shallow clones to speed up build
+variables:
+    GIT_DEPTH: 10
+
+# Keep all jobs within these 3 stages
+stages:
+  - test
+  - build
+  - integ
+
+# This build takes a long time and should be done
+# outside of initial PR CI testing
+openwrt-ar71xx-build:
+  stage: build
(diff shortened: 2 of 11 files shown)


sarcasticadmin:
tux OPENWRT_INT 5-guitarist-flagpole

owendelong:
This is really awesome work, Rob... Thanks... I got the care package from David, but I haven't had a chance to do much with it yet... Hopefully next week.

sarcasticadmin:
@owendelong thanks for reviewing/merging! Keep me posted when you get all setup on your end

PR #379, opened 2021-03-25
[READY] Deduplicate SSH Public Keys

### LINKED ISSUE #359: One source of truth for the SSH keys
## Description
We currently have at least 2 file[s] locations where public ssh keys are stored on thIS GitHub repo:

* In switch configs: https://github.com/socallinuxexpo/scale-network/tree/master/switch-configuration/authentication/keys
* In openwrt: https://github.com/socallinuxexpo/scale-network/blob/master/openwrt/files/root/.ssh/authorized_keys

Is it possible to store this info in one area and have that pulled into whatever configs need it?
This was people who need to up[load, date] keys don't have to do in multi locations.

## Acceptance Criteria
* Centralize ssh publics keys to pull from `facts/` for switch and openwrt configs

kylerisse:
Fixes #359 

## Description of PR
Moves SSH Public Keys into the facts directory and replaces the static openwrt root authorized_keys file with one dynamically generated through gomplate.

## Previous Behavior
Stored SSH Keys in switch configuration directory and had to manually update openwrt root's authorized_keys file manually.

## New Behavior
Update SSH Key in /facts and you're done. OpenWRT golden files have been updated.

## Tests
- verified gomplate template procedure mostly using the test.sh workflow
- kicked off a quick build of openwrt using make to verify no syntax errors with the changes to Makefile, did not wait for it to finish
- executed build_switch_configs.pl to verify functionality


openwrt/files/root/.ssh/authorized_keys
```diff
@@ -1,4 +1 @@
...
-ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBBjjcUJLTENGrV6K/nrPOswcBVMMuS4sLSs0UyTRw8wU87PDUzJz8Ht2SgHqeEQJdRm1+b6iLsx2uKOf+/pU8qE= root@kiev.delong.com
-ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDBP0XP83nOtxnsGJ7AG+FuT7dbAEO8BrIHlPbhCGhKjFKcq8g6LWa0htx3iwHJB2VfasIVSqWAC7SwX81XfSTwpvNSNVXvrIcMNdYOpEBHx5W94iVP2ooVVsGIeK/M+Eu9toiHebtVFT45hQ8VO5DwqbG6xPgEp1PmLsPhyBwP0LyC8Q4tA2ttftC6/fJa6fa1G54c7D/euIhZnyM/ovcpYfNYTlarX/yLJoYaOEk5ByYJPYnpQtm2GQX8CdT/U14nDtJ/VjzW8kP/R6+ZsmGpxBrWp2Mn03LjL4CzTSN3cbE2CAKKSjkIcaaoohRZS+JXnxm5+TXF/6QmN0MPU3SSThKWIzOhzfMrLqOu6hsw16x9lF1eJreUZr/WysWOswM4WNoffBv6mKvVsMR8Z8tgERR8+sIxUaoVOfChR9wRoD5lcNnVxrC6rcF+rBxeY4JiPS3zpyamwteyR0n2YspT9Dl0MlZb4cWml2aBnvYk0TmoCC3eGs+2/wssL4s+64TiFC8WkK7U7RW0eYODwPJQbzNGZBjcsVlKD4whs5cNGlgjkZJzTdFt49EwIKmFnSd9Ly/NY3zhByFS/MN3Mj9I2AqMRZ/i4iT72gATE9aLle1xViGxjK7bHmHT2H1yqQhJ5bMcorpDzSZyTkyBTDDnOgh8tM+tBG204kYywQW/RQ==
+{{ range (datasource "keys_dir") }}{{(file.Read (path.Join (getenv "KEYPATH") .))}}{{end}}
```
kylerisse, line 1: I wanted to have gomplate do the authorized_keys file generation so there wouldn't be seperate steps to maintain in both the Makefile and build.sh. Unfortunately the only way I could get this to work is by using the environment variable KEYPATH to essentially pass the path into gomplate. If you try to interact with the datasource directly then the path is stripped and I couldn't hard code the location because it's different in build.sh and Makefile (one is ../../facts and the other is ../facts). 

I'd love to be able to omit it if there's a way, but this was still cleaner than other options I tried first.

tests/unit/openwrt/golden/ipq806x/root/.ssh/authorized_keys
```diff
@@ -1,4 +1,5 @@
...
 ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDBP0XP83nOtxnsGJ7AG+FuT7dbAEO8BrIHlPbhCGhKjFKcq8g6LWa0htx3iwHJB2VfasIVSqWAC7SwX81XfSTwpvNSNVXvrIcMNdYOpEBHx5W94iVP2ooVVsGIeK/M+Eu9toiHebtVFT45hQ8VO5DwqbG6xPgEp1PmLsPhyBwP0LyC8Q4tA2ttftC6/fJa6fa1G54c7D/euIhZnyM/ovcpYfNYTlarX/yLJoYaOEk5ByYJPYnpQtm2GQX8CdT/U14nDtJ/VjzW8kP/R6+ZsmGpxBrWp2Mn03LjL4CzTSN3cbE2CAKKSjkIcaaoohRZS+JXnxm5+TXF/6QmN0MPU3SSThKWIzOhzfMrLqOu6hsw16x9lF1eJreUZr/WysWOswM4WNoffBv6mKvVsMR8Z8tgERR8+sIxUaoVOfChR9wRoD5lcNnVxrC6rcF+rBxeY4JiPS3zpyamwteyR0n2YspT9Dl0MlZb4cWml2aBnvYk0TmoCC3eGs+2/wssL4s+64TiFC8WkK7U7RW0eYODwPJQbzNGZBjcsVlKD4whs5cNGlgjkZJzTdFt49EwIKmFnSd9Ly/NY3zhByFS/MN3Mj9I2AqMRZ/i4iT72gATE9aLle1xViGxjK7bHmHT2H1yqQhJ5bMcorpDzSZyTkyBTDDnOgh8tM+tBG204kYywQW/RQ==
+ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDrMBEZC9WB9dJMGoev5b+P8I5UtcQLLpoBfKDrEF8bHdQ1AcdpCMtEnTVNmqVMzq8+gRR2CzbqI1Ue7KteFtZX/YtKIEEyL6D5kzhWyQrybl35nxA5VIKJrNszHdXDb/MGMwTMfWV1/VEka4UnCwXpgPH/AXQ/abhmFFDAJnc+PmqXPX3fHA4J5fpl/6CkT7JLKGMpsI7LCtU6cohtnXZrDHqFmPa6T+nHv2cT2QF6eSIPuMX7sLth6QI92yTam8+QUAeMFzEZOcmb6W8CNGio13ezn8ljQ6lB5uZLHhnnuxG9z6RJZugMzU9UxpkHRSKUVX9+qRGf1ntiPKE5Pb4x bibayoff@gmail.com
```
kylerisse, line 5: the automation is already paying off, apparently we already had drift between the two sets of keys. so that's nice.

openwrt/files/root/.ssh/authorized_keys
sarcasticadmin: I follow you. I played around with this a little too and couldnt find a good way to deal with this either. Its weird how `keys_dir` basically gets evaluated into all files in that path but we can find the absolute path of the files themselves (even with other functions). Im good with this and its a straighforward solution 👍 

At some point, I think the easiest thing might be to unify the `Makefile` and `tests/unit/openwrt/test.sh` dir contexts when they execute `gomplate` to generate templates. Then we could hard code the path in `authorized_keys` or at least assume the relative path is the same to the pub keys. A nice side effect of this would be it would remove any fat fingers effect our tests since they would be making the exact same call. I can spin out a follow up for this so look into it further if you think thats a good idea.

sarcasticadmin (approved):
@kylerisse thanks for getting this push up! I think there was one minor change for the openwrt `Makefile` but other than that I can confirm the following were working as expected for me:
* [x] switch configs
* [x] openwrt test.sh

Would you mind updating https://github.com/socallinuxexpo/scale-network/tree/master/switch-configuration#switch-authentication as well? Since its referring to the old path for the switch configs

sarcasticadmin:
Thanks for the fixes to my comments :shipit:

Issue #377, opened 2021-03-17
Consider moving circleci jobs to github actions 


sarcasticadmin:
## Description

Relates to: #373 

We have the majority of the CI jobs running today in circleci. Its been great thus far, but we also might want to consider moving it over to github actions. It would be nice to have a CI that is local to the platform we are running in. Off the top of my head we have a couple of CI jobs that would need to be ported:

* [x] Openwrt golden tests
* [x] Perl Lint and switch-config build tests
* [x] Python unit tests
* [x] Openwrt weekly builds (Covered specifically in #373)

## Acceptance Criteria
* Identify pros and cons of moving
* Prove out equivalent or better functionality in Github Actions

sarcasticadmin:
All looks good on the github actions front, all tests are a good on `master` and in `pull requests`!

![Screenshot at 2022-01-12 20-42-56](https://user-images.githubusercontent.com/30531572/149267413-d728adf1-b6f6-43a9-9d59-42f96666a692.png)

PR #383, opened 2021-07-23
[READY] - Building openwrt imgs using github actions

sarcasticadmin:
## Description of PR

Part of: #377 

As an alternative to our existing circleci builds we current have:
https://github.com/socallinuxexpo/scale-network/blob/a745d0437f1f02da9f290354dda09ab8370eae6b/.circleci/config.yml#L29-L76

There seems to be a good amount of usefulness to doing this in github actions:

1. We get to trigger the builds via `workflow_dispatch` (https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) for any given branch in addition to doing periodic `cron` builds weekly like we current have in circleci. Anyone write access to the repo can now trigger this `workflow_dispatch`. **Example:** if a team member makes a change in a PR, they can trigger this workflow (a build) and then make artifacts for testing. Previously you would have had to do that build locally or after we merge to master with how circleci is currently configured.
1. Authentication is native to github while within the workflows
1. Build times for our images are more or less the same from cicleci what I can tell +-10min (seems to depend on time of day for the github shared runners but there specs are comprebable to circleci: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

If this is approved by the wider team then I think we would leave this in place for a few weeks to see how it compares to circleci. I have some other additional changes coming into to the openwrt image so I can test the build process triggers then.

Eventually these `workflow_dispatch` builds will be able to utilized by the `/tux` command to trigger via pr comments like our tests.

## Previous Behavior
- openwrt image builds only happened via circleci crons

## New Behavior
- openwrt image builds via github actions in addition to existing circleci crons

## Tests
- Tested this in a separate repo (private) per the limitation of some of this needed to be in the default branch to run
Screen Shot 2021-07-23 at 1 24 12 AM

### DIFF
--- a/.github/workflows/openwrt-build.yml
+++ b/.github/workflows/openwrt-build.yml
@@ -1,0 +1,35 @@
+---
+on:
+  workflow_dispatch:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron:  '15 3 * * 0'
+
+jobs:
+  build:
+    name: Building openwrt img
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        target: ["ar71xx", "ipq806x"]
+    container:
+      image: sarcasticadmin/openwrt-build@sha256:bab6de3f66f5365d866de646bb9fd1f2000061d1a5fe8a07b6a714661a9a9a63
+      # Since user is openwrt and gets 1001 from inside container
+      options: --user 1001
+    steps:
+      - uses: actions/checkout@v1
+        with:
+          ref: ${{ env.BRANCH }}
+      - name: Build openwrt
+        run: |
+          cd ${GITHUB_WORKSPACE}/openwrt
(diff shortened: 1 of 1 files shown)


owendelong (approved):

owendelong:
Seems like a good idea to me.

sarcasticadmin:
@owendelong glad you like this as well! Im going to mess around with this a little more but Ill circle back when this is `READY`

sarcasticadmin:
@owendelong this is ready to go! Ill kick off a manual build via the `workflow_dispatch` once this lands it `master`

owendelong:
Great work, Rob. Thanks for doing this!

sarcasticadmin:
@owendelong thanks @owendelong!

PR #395, opened 2021-10-02
[READY] - nix-shell and kermit configs

sarcasticadmin:
## Description of PR
For a discussion and intro of nixpkgs into scale-network

## Previous Behavior
- No kermit config
- No shell.nix

## New Behavior
- Kermit config for junipers
- shell.nix with basic tools

## Tests
How was this PR tested?

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,18 @@
+# TODO: Pin this to a specific version of nixpkgs
+{ pkgs ? import <nixpkgs> {} }:
+
+with pkgs;
+
+let
+  scale_python = python38.withPackages
+      (pythonPackages: with pythonPackages; [ pytest pylint ]);
+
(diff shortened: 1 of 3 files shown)


shell.nix
```diff
@@ -0,0 +1,18 @@
+# TODO: Pin this to a specific version of nixpkgs
```
sarcasticadmin, line 1: Needs a lot of work but this is a decent start :)

kylerisse: One small step for some. One giant leap for us.

kylerisse (approved):
Thanks @sarcasticadmin

shell.nix
sarcasticadmin: hahaha so true :laughing:

PR #416, opened 2022-01-11
[READY] - Update steveb_id_rsa.pub

sbibayoff:
updating my key

## Description of PR
updating my key for use with scale 19x

## Previous Behavior
old key

## New Behavior

## Tests
How was this PR tested?
hopping is works ;-)


sarcasticadmin:
@sbibayoff I went head and helped you out since that change that initially failed was due to the `golden` files for openwrt template testing.

You would normally need to have `gomplate` installed then run:

```
cd openwrt
make golden-update
```

This will regenerate the templates that we tracked for testing purposes.

Should be all set now, but Ill let others on the team review this change as well

sarcasticadmin (approved):

sbibayoff:
Thanks Rob, completely forgot about that. Will keep in mind for future pull requests

sarcasticadmin:
@sbibayoff why did we close this? seems totally fine, going to reopen

kylerisse (approved):
lgtm

Issue #407, opened 2021-10-04
Confirm correct config for dhcp client options


sarcasticadmin:
## Description

dhcp client options definitions:
  - `224` and `225` wifi radio freq: https://github.com/socallinuxexpo/scale-network/blob/master/openwrt/files/etc/udhcpc.user#L8
  - `226` setting network and wireless config file version: https://github.com/socallinuxexpo/scale-network/blob/master/openwrt/files/etc/udhcpc.user#L21

We have options `224` and `225` set for dhclient for openwrt: https://github.com/socallinuxexpo/scale-network/blob/master/facts/secrets/ar71xx-openwrt-example.yaml#L78

However `226` is something we are also keying on but it seems to be missing from the default config in secrets/<arch>-openwrt.yaml

## Acceptance Criteria
- set appropriate `reqopts` on the mgmt interface on APs

PR #422, opened 2022-02-11
[READY] - Adding missing DHCP option 226 to openwrt configs

sarcasticadmin:
## Description of PR
Fixes: #407

Adding missing DHCP options 226 since we need it to trigger the network config switcher script on the APs.

## Previous Behavior
- Option was missing
- Option was temporarily added adhoc after the initial deployment during scale x18

## New Behavior
- DHCP option 226 is being set on the mgmt interface in our default config
- Add documentation about dynamic config
- Additional server spec tests to ensure desired state
- Remove staff wifi from wireless.1 config since that config is really only used for workpartys and init wireless

## Tests
- Openwrt build successful - https://github.com/socallinuxexpo/scale-network/actions/runs/1828060585
- Serverspec tests pass - https://gitlab.com/socallinuxexpo/scale-network/-/jobs/2083565053
  - Includes new tests for things like custom dhcp options


tests/serverspec/spec/shared/openwrt/init.rb
```diff
@@ -107,6 +107,23 @@
       it { should be_grouped_into 'root' }
   end
 
+  describe "correct_num_configs" do
```
kylerisse, line 110: Appreciate the tests

kylerisse (approved):
LGTM, thanks @sarcasticadmin

Issue #388, opened 2021-07-24
Disable password SSH login on AP production images


sarcasticadmin:
## Description

Relates to: #387 

Set `PasswordAuthentication no` instead of the current `PasswordAuthentication yes` for AP production images (images that are going to be used at the conference)

sshd config for APs: https://github.com/socallinuxexpo/scale-network/blob/master/openwrt/files/etc/ssh/sshd_config#L3

## Acceptance Criteria
- Set conditional for disabling SSH password in AP image
- Confirm functionality on actual built AP image
- Add `serverspec` test for checking for password configuration based on prod vs test images

PR #438, opened 2022-05-02
[READY] - Adding option to toggle openwrt sshd PasswordAuth

sarcasticadmin:
## Description of PR

Fixes: #388 

## Previous Behavior
- PasswordAuth for openwrt sshd was always enabled
- scale version in yaml config was 18

## New Behavior
- PasswordAuth for openwrt sshd can be toggled (default to enabled)
- scale version yaml config bumped to 19 for upcoming 19x conference in July

## Tests
- Will do a build and test locally since `autoflash` automation will break when password is disabled


kylerisse (approved):
LGTM

owendelong (approved):
LGTM as well

PR #441, opened 2022-05-30
[READY] - Update udhcpc.user

owendelong:
Fixes mixed-case problem with host name.

## Description of PR
See above

## Previous Behavior
Mixed case host names in config vs. lower case hostnames from KEA didn't compare equally causing erroneous reboots.

## New Behavior
Erroneous reboots from this cause are prevented.

## Tests
/bin/bash

### DIFF
--- a/openwrt/files/etc/udhcpc.user
+++ b/openwrt/files/etc/udhcpc.user
@@ -29,7 +29,7 @@ case "$1" in
     if [ ! -z "$hostname" ]; then
       uci set 'system.@system[0].hostname'="$hostname"
       uci commit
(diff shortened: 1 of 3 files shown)


sarcasticadmin:
Additionally tested this on the netgear APs.

We were running into a scenario where the APs would constantly yoyo between the initial hostname set of `OpenWrt` and what it got form the dhcpd server (kea) `openwrt`. Since hostname case doesnt matter we can just convert everything to the same case (lower) and well above this odd condition caused by the logic in `udhcpc.user`

@owendelong thanks for pushing this up initially. I just rebased and adding in the golden tests for the AP :+1:

sarcasticadmin (approved):

Issue #295, opened 2020-02-24
Current Raspbian Ping doesn't report malformed Echo Replies 


owendelong:
## Description
We don't see echo replies coming back even though they are passed from the kernel to the ping binary.

This is because the uboot firmware on the netgears is sending malformed echo replies.

## Acceptance Criteria
We need a version of ping that still reports these malformed echos.

owendelong:
Strace of current ping ignoring response packets (note the recvmsg system calls:

root@raspberrypi:/home/pi# ping 192.168.1.1 &
[1] 1118
root@raspberrypi:/home/pi# PING 192.168.1.1 (192.168.1.1) 56(84) bytes of data.

root@raspberrypi:/home/pi# strace -fp 1118
strace: Process 1118 attached
gettimeofday({tv_sec=1580619490, tv_usec=379680}, NULL) = 0
poll([{fd=3, events=POLLIN}], 1, 235)   = 0 (Timeout)
gettimeofday({tv_sec=1580619490, tv_usec=615477}, NULL) = 0
gettimeofday({tv_sec=1580619490, tv_usec=615643}, NULL) = 0
sendto(3, "\10\0\vz\4^\0\v\342V6^\333d\t\0\10\t\n\v\f\r\16\17\20\21\22\23\24\25\26\27"..., 64, 0, {sa_family=AF_INET, sin_port=htons(0), sin_addr=inet_addr("192.168.1.1")}, 16) = 64
recvmsg(3, {msg_name={sa_family=AF_INET, sin_port=htons(0), sin_addr=inet_addr("192.168.1.1")}, msg_namelen=128->16, msg_iov=[{iov_base="E\0\0<\0\31\0\0@\1\367R\300\250\1\1\300\250\1\4\0\0P\363\4^\0\vabcd"..., iov_len=192}], msg_iovlen=1, msg_control=[{cmsg_len=20, cmsg_level=SOL_SOCKET, cmsg_type=SCM_TIMESTAMP, cmsg_data={tv_sec=1580619490, tv_usec=615974}}], msg_controllen=20, msg_flags=0}, 0) = 60
recvmsg(3, {msg_namelen=128}, MSG_DONTWAIT) = -1 EAGAIN (Resource temporarily unavailable)
gettimeofday({tv_sec=1580619490, tv_usec=616574}, NULL) = 0
poll([{fd=3, events=POLLIN}], 1, 999)   = 0 (Timeout)
[...]

PR #443, opened 2022-05-30
[READY] - init flash utils and nix-shell pkgs updates

sarcasticadmin:
## Description of PR

Needed to solo flash (old process) 4 APs that didnt take with the massflash. Leveraging the nix-shell as well since the ping util in ubuntu still seem to be broken and doesnt show malformed ping packets :/ see: https://github.com/socallinuxexpo/scale-network/issues/295

## Previous Behavior
- flash util was only workable from the rpi
- no net tools in nix-shell

## New Behavior
- flash util leverages path for vars
- net tools in nix-shell

## Tests
- Flashed the 4 remaining APs

### DIFF
--- a/openwrt/flash
+++ b/openwrt/flash
@@ -1,6 +1,13 @@
 #!/usr/bin/env expect
-spawn /usr/sbin/arp -d 192.168.1.1
-spawn ./ping -n 192.168.1.1
+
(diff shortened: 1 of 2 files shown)


owendelong (approved):
Seems Reasonable

PR #452, opened 2022-07-10
Candidate release for network config for scale 19x

owendelong:
## Description of PR
This should complete the changes necessary to support the network in the Hilton
i(Switches and Routers)

## Previous Behavior
Network was set up for Pasadena Convention Center

## New Behavior
Network is configured for Hilton LAX

## Tests
Not yet tested.


owendelong:
Rob, Please review this carefully and let me know if you see any oversights. Likely some of the data in this configuration change will need to be fed to the Abs.

I based the hi configurations on the ex IP addresses, so few if any v4 changes, but we have a new v6 prefix, so that likely impacts.

Switches 48-50 have been renumbered 51-53 and the new ex4200-48p switches have been numbered 48-50 in the switch types file. Configuration and labeling will have to take place at SCaLE.

owendelong:
Added another major commit to this PR. Looks like I broke something in the process.

owendelong:
Added fix to Makefile to restore "lint" target dependency. Ideally switch the test to depend on the .lint target and remove the lint dependency.

owendelong:
Added fix to Makefile to restore build-switch-configs target as well. Same change to test target is suggested.

router-configuration/to_push/br-mdf-01
```diff
@@ -0,0 +1,701 @@
+## Last changed: 2019-10-27 02:59:16 UTC
```
sarcasticadmin, line 1: date here seems "dated" :laughing:

switch-configuration/config/types/hiRoom
```diff
@@ -1,21 +1,21 @@
...
+TRUNK	ge-0/0/13	hiInfra,hiSCALE-SLOW,hiSCALE-FAST,hiStaff				// AP
+TRUNK	ge-0/0/14	hiInfra,hiSCALE-SLOW,hiSCALE-FAST,hiStaff				// AP
+TRUNK	ge-0/0/15	hiInfra,hiSCALE-SLOW,hiSCALE-FAST,hiStaff				// AP
+TRUNK	ge-0/0/16	hiInfra,hiSCALE-SLOW,hiSCALE-FAST,hiStaff				// AP
+TRUNK	ge-0/0/17	hiInfra,hiSCALE-SLOW,hiSCALE-FAST,hiStaff				// AP
+VLAN	hiSigns		4	// ge-0/0/{18-21}
+VLAN	hiAVLAN		6	// ge-0/0/{22-27}
+VLAN	hiSpeaker	2	// ge-0/0/{28,29}
+VLAN	hiCTF		1	// ge-0/0/30
+RSRVD	1			// ge-0/0/31 (was HAM_BRIDGE)
```
sarcasticadmin, line 19: Thanks for cleaning up the HAM bridge. I know that not be required from a few scales now IIRC

switch-configuration/config/switchtypes
```diff
@@ -13,57 +13,57 @@
...
-SanLorenzoA	45	503	2001:470:f325:503::200:45	hiRoom			D.9		Normal
-SanLorenzoE	46	503	2001:470:f325:503::200:46	hiRoom			D.9		Loud
-SanLorenzoF	47	503	2001:470:f325:503::200:47	hiRoom			D.9		Normal
-TTLB-LBY	48	503	2001:470:f325:503::200:48	hiIDF			M.2		??
-TT2E-2FCat	49	503	2001:470:f325:503::200:49	hiIDF			M.2		??
-TT2D-2FReg	50	503	2001:470:f325:503::200:50	hiIDF			M.2		??
-SpareA		51	503	2001:470:f325:503::200:51	hiIDF			Z.9		??
-SpareB		52	503	2001:470:f325:503::200:52	hiIDF			Z.9		??
-SpareC		53	503	2001:470:f325:503::200:53	hiIDF			Z.9		??
+//Name		Num	MgtVL	IPv6				Type		Hierarchy	Noiselevel	Model
```
sarcasticadmin, line 16: Its nice to see that youve got the models in here now. Im also glad weve been able to standardize by and large on the ex4200-48p

switch-configuration/config/vlans
```diff
@@ -1,9 +1,11 @@
...
 
 // Conference Center -- VLANS 500-899
-#include vlans.d/Conference
+//#include vlans.d/Conference
 
 // Bridged -- VLANS 900-999 (carried across fiber between buildings)
-#include vlans.d/Bridged
+//#include vlans.d/Bridged
 
+// Hilton -- SCaLE 19x special
```
sarcasticadmin, line 10: With `vlans.d` does this mean that we actually have a holistic view for both pasadena and lax hilton? I know we wont be at the hilton past this year but it does seem promising that we have the ability to include additional venues if/when the time comes via this approach. And all you have to do is comment out which of the `includes` you dont want in the future.

switch-configuration/config/vlans.d/Hilton
```diff
@@ -0,0 +1,18 @@
+//	Conference	Center	--	VLANS	100-499
+VLAN	hiSCALE-SLOW		100	2001:470:f0fb:100::/64	10.128.128.0/21	2.4G Wireless Network in
```
sarcasticadmin, line 2: vlan allocations for hilton look good, not that its a show stopper but any particular reason you went with 128 for the second octet for ipv4?

switch-configuration/config/vlans.d/Hilton
```diff
@@ -0,0 +1,18 @@
...
+VLAN	hiSCALE-SLOW		100	2001:470:f0fb:100::/64	10.128.128.0/21	2.4G Wireless Network in 
+VLAN	hiSCALE-FAST		101	2001:470:f0fb:101::/64	10.128.136.0/21	5G Wireless Network in 
+VLAN	hiSpeaker		102	2001:470:f0fb:102::/64	10.128.2.0/24	Speaker Network
+VLAN	hiInfra			103	2001:470:f0fb:103::/64	10.128.3.0/24	SCALE Network Infrastructure and Servers
+VLAN	hiCTF			104	2001:470:f0fb:104::/64	10.128.4.0/24	Capture the Flag
+VLAN	hiAVLAN			105	2001:470:f0fb:105::/64	10.128.5.0/24	Audio Visual Network (DHCP Helper to AV server)
+VLAN	hiNOC			106	2001:470:f0fb:106::/64	10.128.6.0/24	Wired network in NOC
+VLAN	hiSigns			107	2001:470:f0fb:107::/64	0.0.0.0/0	Signs network (IPv6 Only)
+VLAN	hiStaff			108	2001:470:f0fb:108::/64	10.128.8.0/24	Staff Wifi Network
+VLAN	hiInstall		110	2001:470:f0fb:110::/64	10.128.10.0/24	Install Fests and Workshops
```
sarcasticadmin, line 11: is there a reason we've swapped out the vlan 110? Orginally it was reg and now it looks like its installs (install fests, etc.). Do we need to factor in reg this time around or what are your thoughts there?

ExReg (vlans.d/Expo)
```
VLAN    exRegistration          110     2001:470:f325:110::/64  10.0.10.0/24    Registration Network (Expo Center) IPv6 Only
```

HiInstall (vlans.d/Hilton)
```
VLAN    hiInstall               110     2001:470:f0fb:110::/64  10.128.10.0/24  Install Fests and Workshops
```

switch-configuration/Makefile
```diff
@@ -6,20 +6,37 @@ ifndef PERL
...
-lint:
-	find . -iname "*.pl" | xargs -i perl -Mstrict -Mdiagnostics -M'warnings FATAL=>q(all)' -cw {}
+all: switch-maps-bundle
 
-secrets:
+lint:	.lint
+
+.lint: config/scripts/*.pl
+	find . -name '*.pl' | xargs '-I{}' perl -Mstrict -Mdiagnostics -M'warnings FATAL=>q(all)' -cw {}
+	touch .lint
```
sarcasticadmin, line 15: Nice approach to making the makefile dependencies be honored and trigger only when changes to the target actually have updates :+1:

router-configuration/to_push/br-mdf-01
```diff
@@ -0,0 +1,701 @@
...
+                address 172.20.0.1/24;
+            }
+            family inet6 {
+                address 2001:470:f0fb:8000::1/64;
+            }
+        }
+        unit 999 {
+            description "Internet Link to Hilton/AT&T";
+            family inet {
+                address 12.203.68.60/23;
```
sarcasticadmin, line 365: new IP for the hilton, looks like it matches your notes on the mailing list

sarcasticadmin (approved):
@owendelong Overall this looks great to me! Thank you for correcting the CI test failures as well, I think those changes related to the makefile seem reasonable to me.

The stickers that you attached in the mailing list looks great too!

router-configuration/to_push/br-mdf-01
owendelong: Yeah, it is, it's just a comment that was generated when the config I used as a template was saved from the router. I can delete that line if you care.

switch-configuration/config/types/hiRoom
owendelong: It's still there in the PCC config files, IIRC, but it makes no sense in the Hilton context since there aren't any separate buildings via routed segments.

switch-configuration/config/vlans
owendelong: Yes... vlans.d essentially contains files included by directives in the vlans file in the parent directory. By turning those on and off, we can include or not include various files in the vlans.d directory. However, currently, we couldn't include both because the Hilton VLANs and the CF VLANs would collide.

switch-configuration/config/vlans.d/Hilton
owendelong: That's an error... It should be 10.0... Good catch.

switch-configuration/config/vlans.d/Hilton
owendelong: Because even though we're using EX VLAN numbers and EX network addresses, most of the Hilton is actually templated from CF files which were closer to what we needed. Registration is unique to EX and so it got added back in. Likely Install was previously 410.

switch-configuration/Makefile
owendelong: Glad you like it.

router-configuration/to_push/br-mdf-01
owendelong: Correct.

owendelong:
So on the changes to the Makefile and the CI test consequences... How do you feel about updating the tests to use the new targets which produce ".<name>" files to mark dependency state instead of requiring the additional target names just for the CI tests?

switch-configuration/config/switchtypes
owendelong: Yeah, all columns after "Type" don't really have any meaning to most of the code (Though Hierarchy is used by the auto pinger script(s) and may get used later by some of the deployment scripts). Adding the model numbers seemed to make sense and standardizing was a goal from very early on in my tenure on the tech team after all the fun we had with no 2 HP switches taking the sam configuration syntax.

router-configuration/to_push/br-mdf-01
owendelong: I've deleted the line in question.

switch-configuration/config/vlans.d/Hilton
owendelong: I've fixed that, now 10.0.x.x.

owendelong:
The following changes are about to be committed to resolve the above described problems. Good catches, Rob!! Thanks!

switch-configuration/config/vlans
sarcasticadmin: totally makes sense

PR #458, opened 2022-07-26
[READY] - Owen pre 19x.06

owendelong:
## Description of PR
SCaLE 19x as built for initial deploy.

## Previous Behavior
Not yet deployed.

## New Behavior
All switches have been config loaded and scripts verified to work.

## Tests
We actually built the switch configurations.


sarcasticadmin:
Thanks for the detailed commit messages @owendelong 

Looks good and confirmed that junos configs are as we expect :shipit:

sarcasticadmin (approved):

PR #461, opened 2022-07-26
[READY] - Updates to the core role for ansible during 19x deployment

sarcasticadmin:
## Description of PR

Updates to the `core` role for ansible during 19x deployment

## Previous Behavior
- Support for 2 dns servers in dhcpd
- named acl conditional led to inconsistent templating
- Bind dnssec set to `auto` (default from bind iirc)

## New Behavior
- Support for 1 or 2 dns server in dhcpd
- named acl conditional led to inconsistent templating
- Bind DNSSEC set to `no` - disabling DNSSEC for the root recursive name servers 
- Updates to the serverlist.csv for `core1` and updates to the ipv6 prefix

## Tests
- ansible-playbook -u ubuntu -i inventory.py core.yml
> `core.yml` is a temp file to only test the single role


owendelong (approved):
Looks good. Thanks!

PR #465, opened 2022-07-27
[READY] - Missing character on AP serial

ptrlow1:
## Description of PR
Added a missing character that's missing on an AP serial


sarcasticadmin (approved):
@ptrlow1 thanks for catching this during the workparty! :shipit:

PR #482, opened 2022-07-28
Updates to switchtypes to support Hilton Expo (19x)

owendelong:
## Description of PR
Configure switches for Expo floor at Hilton (19x)

## Previous Behavior
Incorrect/insufficient switch setup for 19x

## New Behavior
This should be correct until it isn't.

## Tests
It's just switches... What could possibly go wrong?


sarcasticadmin (approved):
@owendelong looks good, thanks for the updates!

PR #484, opened 2022-07-28
use set; and in returns boolean

JimDennis:
## Description of PR
In function isvalidmodel(model) we can use a set for the enumeration, and the in operator returns a boolean (so just return model in ...)

## Previous Behavior
Style/brevity change; no change in behavior

## Tests
Tested in REPL


owendelong (approved):
I fucking hate python

PR #485, opened 2022-07-28
fix bad IP subnet

davidelang:
fix bad IP subnet

script put 10.3.0 instead of 10.0.3, explaining why the reservations were ignored. This should fix them.

sorry for not making a new branch for it.


owendelong (approved):
That looks much more better now.

PR #491, opened 2022-07-29
fix ap mac addresses

davidelang:
fixing mac addresses so everything gets proper dhcp info.


sarcasticadmin (approved):
@davidelang thank you for fixing this up!

PR #493, opened 2022-07-29
add csv header

davidelang:
it appears that the conversion script that uses aplist throws away the first line of the file, so adding a header line


sarcasticadmin:
:+1:

sarcasticadmin (approved):

PR #502, opened 2022-07-30
initial version of script to make topo from lldp

davidelang:
script that scrapes switch configs for lldp data and creates png and svg diagrams


owendelong (approved):
Great first steps!!

PR #507, opened 2022-10-11
OOPS -- Wrong type specified for NOC switch, corrected.

owendelong:
## Description of PR
Correct type spec for NOC switch

## Previous Behavior
Invalid type name

## New Behavior
Valid PCC type name

## Tests
Tests -- We don't need no steenkin' tests.


owendelong:
This is such a trivial change and it is causing tests to fail for other PRs, so I am going to go ahead and override and merge without review.

Issue #516, opened 2022-12-28
Applying for gitlab's open source program


sarcasticadmin:
## Description

It looks like gitlab's free tier changed sometime mid year-2022 and thus we are no longer to have the mirror at gitlab pull from this repo in gitlab. This results in being unable to leverage the gitlab runners for testing openwrt seamlessly. We can apply for the open source tier which grants us access to the features we need to continue using the mirror inside of gitlab. This requires we set a proper OSS license as well in the repo.

More info: https://about.gitlab.com/solutions/open-source/join/

## Acceptance Criteria
- Add LICENSE to scale-network repo
- Apply for access to the program

sarcasticadmin:
Submission has been made 
![ss-202212281672247696](https://user-images.githubusercontent.com/30531572/209848653-faaa758e-dd74-44b8-9f1d-ffad515dff4b.png)

sarcasticadmin:
We got approved! Ive enabled the open source tier and its good for the next year. It looks like we have to renew annually.

![ss-202301241674601320](https://user-images.githubusercontent.com/30531572/214515613-27edbcb5-1c15-46c6-a652-2fdadfe16e68.png)

PR #517, opened 2022-12-28
[READY] - init BSD 3-clause license

sarcasticadmin:
## Description of PR

Relates to: #516 

After speaking with @owendelong we've decided on the following license for this codebase.

## Previous Behavior
- No license was present in the repo

## New Behavior
- License set to BSD 3-clause

## Tests
- NA


owendelong (approved):
This is both the path  of least resistance and the right thing to do.

I approve this message.

PR #526, opened 2023-01-29
owen 20x 09

owendelong:
- Recreated minimal switch config "miniconfig" Bug fixes for bulk_local_load_switches
- Updated switchtypes with correct MAC addresses for 24 port switches
- Created template for mass flashing


switch-configuration/config/types/MassFlash
```diff
@@ -0,0 +1,49 @@
+// Mass Flashing Switch Configuration
```
sarcasticadmin, line 1: all the ports!

sarcasticadmin (approved):

PR #530, opened 2023-01-29
[READY] - openwrt configs pre show

sarcasticadmin:
## Description of PR
- Updating openwrt mods for SHA512 hashs in `/etc/shadow`
- Removing conntrack modules


owendelong (approved):
Let's flash!

PR #531, opened 2023-01-30
owen 20x 11

owendelong:
- Improved switch_pinger to skip hosts down hierarchy from down hosts Fixed daisy chain for CTF switches in switchtypes
- Added POE configuration capability (PTRUNK hack). Improved security of miniconf


sarcasticadmin (approved):
Thanks for clarifying this with me, looks good!

PR #534, opened 2023-02-08
[REVIEW] - init bhyve image support for nixos images

sarcasticadmin:
## Description of PR

Adding necessary bits to support running on bhyve for the images produced via nix.

## Previous Behavior

- loghost was only being specific to a qemu image format

## New Behavior

- update tests to use `callPackage`
- adding bhyve config for loghost - Proposing to put this in `utilties/bhyve` instead of `ansible/roles/bhyve`
- bhyve image nix function
- loghost support bhyve image gen

## Tests

Able to build and run a nixos vm on bhyve:

```
$ nix build ".#nixosConfigurations.x86_64-linux.loghost.config.system.build.bhyveImage"
```
> Produces a bootable, compatible image when applied with the vm config

Building qemu images still works as well:

```
$ nix build ".#nixosConfigurations.x86_64-linux.loghost.config.system.build.vm"
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -31,7 +31,7 @@
           common =
             ({ modulesPath, ... }: {
               imports = [
-                "${toString modulesPath}/virtualisation/qemu-vm.nix"
+                ./nix/modules/bhyve-image.nix
               ];
             });
           pkgs = nixpkgsFor.${system};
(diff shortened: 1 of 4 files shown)


owendelong (approved):
Great work, Rob! Looking forward to this.

PR #538, opened 2023-03-01
[READY] - Updates to core server and kea config

sarcasticadmin:
## Description of PR

Updates to kea configuration on core server.

## Previous Behavior
- Had issues with mac address setting via systemd

## New Behavior
- simplify the systemd-networkd config (going to override the mac at the bhyve level instead)
- Adding actual client/server dhcpd test for core
- define user-context per subnet for vlan name
- init subnet ids for kea
- core test wait for network on coreserver to common online

## Tests

Client/server test for dhcp passes. Current test scenario is: Kea server starts with currrent configuration and is tested via builtin `-t` flag. Kea client comes up and tries to get a lease then pings the server:

```
nix build ".#packages.x86_64-linux.scaleTests.core"
```

Confirmed dhcp options are being passed as expected with APs configurations as well.

### DIFF
--- a/facts/inventory.py
+++ b/facts/inventory.py
@@ -556,11 +556,13 @@ def main():
         for ap in aps
     ]
 
-    # TODO: support generating id for each subnet block
-    # called out in: https://kea.readthedocs.io/en/latest/arm/dhcp4-srv.html#ipv4-subnet-identifier
     subnets_dict = [
         {
             "subnet": vlan["ipv4prefix"] + "/" + str(vlan["ipv4bitmask"]),
+            # generating uniq id (prefix with dots) for each subnet block to ensure autoids dont effect reordering
+            # called out in: https://kea.readthedocs.io/en/latest/arm/dhcp4-srv.html#ipv4-subnet-identifier
+            # subnet ids must be greater than zero and less than 4294967295
+            "id": int(vlan["ipv4prefix"].replace('.', '')),
             "user-context": { "vlan": vlan["name"] },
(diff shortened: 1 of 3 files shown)


facts/inventory.py
```diff
@@ -556,11 +556,14 @@ def main():
...
 
-    # TODO: support generating id for each subnet block
-    # called out in: https://kea.readthedocs.io/en/latest/arm/dhcp4-srv.html#ipv4-subnet-identifier
     subnets_dict = [
         {
             "subnet": vlan["ipv4prefix"] + "/" + str(vlan["ipv4bitmask"]),
+            # generating uniq id (prefix with dots) for each subnet block to ensure autoids dont effect reordering
+            # called out in: https://kea.readthedocs.io/en/latest/arm/dhcp4-srv.html#ipv4-subnet-identifier
+            # subnet ids must be greater than zero and less than 4294967295
+            "id": int(vlan["ipv4prefix"].replace('.', '')),
```
sarcasticadmin, line 565: I figured this was unique enough for our use case. But let me know if you have other ideas how to generate an appropriate id @owendelong

facts/inventory.py
```diff
@@ -556,11 +556,14 @@ def main():
...
+            "user-context": { "vlan": vlan["name"] },
```
sarcasticadmin, line 566: user-context is just arbitrary data that we define, I figure this was at least better than nothing

owendelong (approved):
This looks awesome... Thanks, Rob!!

PR #543, opened 2023-03-07
[READY] - supporting bind master and slave with nvme devices 

sarcasticadmin:
## Description of PR

Support bind slave for coreconf vm

## Previous Behavior
- Originally vms were only bootable via ahci device type
- no slaves for bind were supported

## New Behavior
- split master & slave configs for core machine
- adding common to core
- support nvme devices for booting
- common config cleanup

## Tests

Will test completely when hosts are deployed

Tests passing locally

```
$ nix build ".#packages.x86_64-linux.scaleTests.core"
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -56,11 +56,18 @@
               ./nix/machines/massflash.nix
             ];
           };
-          core = nixpkgs.lib.nixosSystem {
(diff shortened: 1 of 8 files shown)


owendelong (approved):
Um, what's this button do?

PR #544, opened 2023-03-07
Added PoE capability for access ports and improved PoE configuration mechanism

owendelong:


sarcasticadmin (approved):
@owendelong looks great, nice work!

PR #547, opened 2023-03-08
added git and vim to coreMaster coreSlave

nixinator:
## Description of PR
add git and vim to coreMaster and slaveMaster

## Tests

git clone git@github.com:socallinuxexpo/scale-network.git

cd scale-network

nix repl

>>:lf .
>>:b nixosConfigurations.x86_64-linux.coreMaster.config.system.build.toplevel
>>:b nixosConfigurations.x86_64-linux.coreSlave.config.system.build.toplevel

### DIFF
--- a/nix/machines/core/common.nix
+++ b/nix/machines/core/common.nix
@@ -23,6 +23,8 @@
     bind
     kea
     scaleInventory
+    vim
+    git
   ];
 
   environment.etc."bind/named.conf".source = config.services.bind.configFile;


nix/machines/core/common.nix
```diff
@@ -23,6 +23,8 @@
     bind
     kea
     scaleInventory
+    vim
```
sarcasticadmin, line 26: Finally!

sarcasticadmin (approved):

nix/machines/core/common.nix
nixinator: You can half my payments.

if only you could divided by zero?

PR #559, opened 2023-03-10
Removed unused switches from switch_pinger groups

owendelong:


genebean (approved):
This seems reasonable to me based on my limited understanding of the network steup

sarcasticadmin (approved):

PR #562, opened 2023-03-10
Fixed ExpoA1

owendelong:


sarcasticadmin (approved):
@owendelong good catch! :shipit:

Issue #582, opened 2023-03-11
make recipes for common nix build tasks


sarcasticadmin:
## Description

We have the following common build tasks for this repo:

- building bhyveImages: ` nix build ".#nixosConfigurations.<machine>.config.system.build.bhyveImage"`
- building a qemu vm: `nix build ".#nixosConfigurations.<machine>.config.system.build.vm"`
- running tests found nix `nix/tests`: `nix build ".#packages.x86_64-linux.scaleTests.<test>"`

## Acceptance Criteria
- Having make recipes to simplify the build processes for nix tasks

sarcasticadmin:
completed via #598 Thanks @nixinator !

PR #598, opened 2023-03-24
makefile wrapper for issue #582

nixinator:
## Description of PR
Added a make file for common nix build targets #582

## New Behavior
make show
make bhyveImage signs  (this will make the signs server)

We can test this in CI i suppose?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,36 @@
+Command := $(firstword $(MAKECMDGOALS))
+Arguments := $(wordlist 2,$(words $(MAKECMDGOALS)),$(MAKECMDGOALS))
+
+default: show 
+
+bhyveImage:
+	@echo "nix building $(Arguments) bhyve Image"
+	nix build ".#nixosConfigurations.$(Arguments).config.system.build.bhyveImage"
+
+qemuVm:
+	@echo "nix building $(Arguments) qemu VM"
+	nix build ".#nixosConfigurations.$(Arguments).config.system.build.vm"
+
+all:	bhyveImage qemuvm
+        	
+
+show:
+	@echo "A simple wrapper for nix build "$(Arguments)"!";
+	@echo "You will need NIXOS Installed!!! :-)";
+	@echo "Available test are:"
+	@echo 
+	@ls nix/tests/.
+	@echo 
+	@echo "Try typing make at the CLI and hit TAB (on your keyboard) for targets"
+	@echo 
+	nix flake show;
+
+nixtests: 
+	@echo "Available test are:"
+	@ls nix/tests/.
+	@echo
+	@echo "nix building and running nixos $(Arguments) tests"
+	nix build ".#packages.x86_64-linux.scaleTests.$(Arguments)"
+
+%::
+	@true
--- a/Makefile
+++ b/Makefile
@@ -5,13 +5,13 @@ default: show
 
 bhyveImage:
 	@echo "nix building $(Arguments) bhyve Image"
-	nix build ".#nixosConfigurations.$(Arguments).config.system.build.bhyveImage"
+	nix build ".#nixosConfigurations.$(Arguments).config.system.build.bhyveImage" -o result-bhyveImage-$(Arguments)
 
 qemuVm:
 	@echo "nix building $(Arguments) qemu VM"
-	nix build ".#nixosConfigurations.$(Arguments).config.system.build.vm"
+	nix build ".#nixosConfigurations.$(Arguments).config.system.build.vm" -o result-qemuVm-$(Arguments)
 
-all:	bhyveImage qemuvm
+all:	bhyveImage qemuVm
         	
 
 show:


Makefile
```diff
@@ -0,0 +1,36 @@
...
+	nix flake show;
+
+nixtests: 
+	@echo "Available test are:"
+	@ls nix/tests/.
+	@echo
+	@echo "nix building and running nixos $(Arguments) tests"
+	nix build ".#packages.x86_64-linux.scaleTests.$(Arguments)"
+
+%::
```
sarcasticadmin, line 35: Whats this shorthand do in this case? my makefu is rusty

Makefile
```diff
@@ -0,0 +1,36 @@
...
+	nix build ".#nixosConfigurations.$(Arguments).config.system.build.bhyveImage"
+
+qemuVm:
+	@echo "nix building $(Arguments) qemu VM"
+	nix build ".#nixosConfigurations.$(Arguments).config.system.build.vm"
+
+all:	bhyveImage qemuvm
+        	
+
+show:
```
sarcasticadmin, line 17: Nice I like this addition! Actually caused me to spin out a follow up issue (https://github.com/socallinuxexpo/scale-network/issues/600) since it looks like I have to do something a little more cleaver for the package scope for scaleTests in the flake.nix

Makefile
```diff
@@ -0,0 +1,36 @@
+Command := $(firstword $(MAKECMDGOALS))
```
sarcasticadmin, line 1: Simple way to parse args, nice

Makefile
```diff
@@ -0,0 +1,36 @@
...
+
+bhyveImage:
+	@echo "nix building $(Arguments) bhyve Image"
+	nix build ".#nixosConfigurations.$(Arguments).config.system.build.bhyveImage"
+
+qemuVm:
+	@echo "nix building $(Arguments) qemu VM"
+	nix build ".#nixosConfigurations.$(Arguments).config.system.build.vm"
+
+all:	bhyveImage qemuvm
```
sarcasticadmin, line 14: This should be `qemuVm`

Also, if we built both result will be overriden in the cwd which Im ok with but maybe its worth adding a warning to the user since they might expect `result`, `result2`, etc.

sarcasticadmin (commented):
@nixinator thanks for pushing this up! Testing locally is sufficient for now :+1: I had a couple of comments but this is looking really nice!

Confirming the functionality myself as well:

```
$ make
A simple wrapper for nix build !
You will need NIXOS Installed!!! :-)
Available test are:

allTests.nix
core.nix
loghost.nix
result

Try typing make at the CLI and hit TAB (on your keyboard) for targets

nix flake show;
git+file:///home/rherna/workspace/scale-network?ref=refs%2fheads%2fadd-makefile&rev=e194d74ec0ddc4116db4459276c7c49b4b289a82
├───devShells
│   └───x86_64-linux
│       └───default: development environment 'nix-shell'
├───nixosConfigurations
│   ├───coreMaster: NixOS configuration
│   ├───coreSlave: NixOS configuration
│   ├───loghost: NixOS configuration
│   ├───massflash: NixOS configuration
│   └───signs: NixOS configuration
├───overlays
│   └───default: Nixpkgs overlay
└───packages
    └───x86_64-linux
        ├───scaleInventory: package 'scaleInventory'
```

Building a bhyveImage looks good too:

```
$ make bhyveImage coreMaster
nix building coreMaster bhyve Image
nix build ".#nixosConfigurations.coreMaster.config.system.build.bhyveImage"
```

Makefile
nixinator: oh, i was experiment with something, but it's spurious.. and should not be there!

Makefile
nixinator: oh yeah, good catch, i can probably change the results symlink name to some different, but didn't because nixxers tend to build one things an expect it to be place in result.

Makefile
sarcasticadmin: :laughing: you missed your opportunity to make me believe some wild new capability of make

Makefile
sarcasticadmin: Im fine with keeping to the convention of result and if I want something I can run a subsequent command to get `result` linked to the store.

Makefile
nixinator: i think i needed it catch make trying to 'build signs' as a target.. .let me investigate.. make really does show it's age.

Makefile
nixinator: I've put them into different symlink outputs, so there should be no confusion now.

Makefile
```diff
@@ -0,0 +1,36 @@
+Command := $(firstword $(MAKECMDGOALS))
+Arguments := $(wordlist 2,$(words $(MAKECMDGOALS)),$(MAKECMDGOALS))
+
+default: show 
+
+bhyveImage:
+	@echo "nix building $(Arguments) bhyve Image"
+	nix build ".#nixosConfigurations.$(Arguments).config.system.build.bhyveImage" -o result-bhyveImage-$(Arguments)
```
sarcasticadmin, line 8: cool, this works for me

sarcasticadmin (approved):
@nixinator Thanks so much! :shipit:

Issue #511, opened 2022-12-24
Convert /tux keyword action to use standardize-pr-comment


sarcasticadmin:
## Description

Currently we are only able to test changes to https://github.com/socallinuxexpo/scale-network/blob/master/.github/workflows/trigger.yml after it lands in `master` due to the way the action triggers out of the box from github actions.

The upstream project: https://github.com/Nebulaworks/action-standardize-pr-comment provides a way of unifying the comment apis provided by actions so that PRs that effect the comment APIs inside the action can be tested before landing in `master`.

## Acceptance Criteria
- trigger.yml action now has the ability to be triggered during an open PR

owendelong:
Yes, please!!

Owen

PR #606, opened 2023-04-15
[READY] - update workflows runs-on to 22.04

sarcasticadmin:
## Description of PR

Fixes: #605

Some of the GHA workflows were using a deprecated version of ubuntu and
it was causing failures. Example:
https://github.com/socallinuxexpo/scale-network/actions/runs/4648537091

## Previous Behavior
- Some workflows were running on 18.04 which was deprecated

## New Behavior
- Bumped workflows to use 22.04

## Tests

- Able to trigger a workflow via this branch and it built successfully: https://github.com/socallinuxexpo/scale-network/actions/runs/4706054646

PR #607, opened 2023-04-15
[READY] - init pr-standardize-comment for trigger workflow

sarcasticadmin:
## Description of PR

Fixes: #511

Depends on: #606 

Leverage the opensource workflow: https://github.com/Nebulaworks/action-standardize-pr-comment

Allows us to be able to test the workflows while in the PR vs having to merge and then test (fail on continuous delivery).

## Previous Behavior
- Could only test the trigger GHA workflow on `master`

## New Behavior
- Can test the `trigger` GHA workflow from PR

## Tests
- I believe this will need to be merged in `master` prior to being able to test this for the first time

### DIFF
--- a/.github/workflows/trigger.yml
+++ b/.github/workflows/trigger.yml
@@ -15,12 +15,15 @@ jobs:
       && github.event.issue.pull_request
       && github.actor == 'sarcasticadmin'
     steps:
+    - name: Standardize PR comment
+      id: pr-standardize-comment
+      uses: nebulaworks/action-standardize-pr-comment@v1.0
     - name: 'Collect event data'
       id: job
       run: |
         set -eux
-        SUBCOMM=$(cut -d ' ' -f 2 <<< "${{ github.event.comment.body }}")
-        SUBARGS="$(cut -d ' ' -f 3- <<< "${{ github.event.comment.body }}")"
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger.yml
```diff
@@ -15,12 +15,15 @@ jobs:
       && github.event.issue.pull_request
       && github.actor == 'sarcasticadmin'
     steps:
+    - name: Standardize PR comment
```
sarcasticadmin, line 18: /tux openwrt build run-at-test

.github/workflows/trigger.yml
```diff
@@ -15,12 +15,15 @@ jobs:
...
+      id: pr-standardize-comment
+      uses: nebulaworks/action-standardize-pr-comment@v1.0
```
sarcasticadmin, line 20: /tux openwrt build run-at-test

kylerisse (approved):
I believe you are correct @sarcasticadmin , needs to be merged to test. Looks right to me. Worst case something borks and we follow up PR it I suppose...

sarcasticadmin:
Thanks @kylerisse I moved it to `[READY]` and then well see what happens :shipit:

Issue #367, opened 2020-06-24
Update openwrt network configs for openwrt dsa conversion


sarcasticadmin:
## Description
@davidelang caught this on the mailing list

~~When we upgrade to openwrt 20.0x its been suggested we update our network configurations for the onboard interfaces and switches within the AP hardware.~~ WNDR 3700v2 and similar devices do not support dsa at this time.

This originated from the mailing list: http://lists.infradead.org/mailman/listinfo/openwrt-adm

Will get more specifics when Im fulled added to this.

Update:
- Its referred to as `dsa`, see: https://openwrt.org/docs/guide-user/network/dsa/converting-to-dsa

## Acceptance Criteria
* Update the switch configuration for openwrt dsa conversion

sarcasticadmin:
So attempted to take a stab at this today... Heres what I found:

sarcasticadmin:
Good news is the potential front runner for https://github.com/socallinuxexpo/scale-network/issues/410 - https://openwrt.org/toh/linksys/e8450 has a mediatek chipset with `DSA` support already in `master`: https://github.com/openwrt/openwrt/blob/9260027535b8f7eab5481b7f545a3fc6477d14fa/target/linux/mediatek/mt7622/config-5.15#L278-L280

So what we might be able to do is just conditionally support configuration for `DSA` (/etc/config/network) in the single architecture until the Netgear WNDR 3700v2, 3800, 3800ch have DSA support

sarcasticadmin:
Discussion about openwrt dsa bridge with mixed vlans:

https://forum.openwrt.org/t/migrating-to-dsa-bridging-across-vlans-not-ports/138061
https://forum.openwrt.org/t/dsa-with-two-vlan-on-one-network/111909

Issue #532, opened 2023-01-30
Belkin RT3200 OpenWrt conversion 


ptrlow1:
## Description
As we intend to use the RT3200s to run OpenWrt from now on, we need to convert these to "UBI" layout before flashing the image.

https://openwrt.org/toh/linksys/e8450#installation

## Acceptance Criteria
Convert to UBI layout by using this source from this git repo: 
https://github.com/dangowrt/linksys-e8450-openwrt-installer

ptrlow1:
Note that this is an inital flash. (We **only** do this **once**.) After that, it will be part of the auto AP mass flash operation we do at work parties.

ptrlow1:
Finished converting the 3200s to OpenWrt late Wednesday night (right before they threw us out). Rob will finish those with our built image, and will be deployed to the show.

ptrlow1:
Images built by Rob didn't work as they were built without "UBI" support. 

https://openwrt.org/docs/techref/flash.layout#ubi_unsorted_block_images

sarcasticadmin:
Posting some stock configs for reference

/etc/config/wireless:
```

config wifi-device 'radio0'
	option type 'mac80211'
	option path 'platform/18000000.wmac'
	option channel '1'
	option band '2g'
	option htmode 'HT20'
	option disabled '1'

config wifi-iface 'default_radio0'
	option device 'radio0'
	option network 'lan'
	option mode 'ap'
	option ssid 'OpenWrt'
	option encryption 'none'

config wifi-device 'radio1'
	option type 'mac80211'
	option path '1a143000.pcie/pci0000:00/0000:00:00.0/0000:01:00.0'
	option channel '36'
	option band '5g'
	option htmode 'HE80'
	option disabled '1'

config wifi-iface 'default_radio1'
	option device 'radio1'
	option network 'lan'
	option mode 'ap'
	option ssid 'OpenWrt'
	option encryption 'none'
```

/etc/config/network:

```

config interface 'loopback'
	option device 'lo'
	option proto 'static'
	option ipaddr '127.0.0.1'
	option netmask '255.0.0.0'

config globals 'globals'
	option ula_prefix 'fda4:b995:95b4::/48'

config device
	option name 'br-lan'
	option type 'bridge'
	list ports 'lan1'
	list ports 'lan2'
	list ports 'lan3'
	list ports 'lan4'

config interface 'lan'
	option device 'br-lan'
	option proto 'static'
[...]

PR #611, opened 2023-08-08
[READY] - Belkin RT3200 support for scale openwrt

sarcasticadmin:
## Description of PR
Fixes: #367 #532 

I wanted to get this to a place where it was workable before it got any larger. But this adds support for a new piece of hardware: Belkin RT3200/linksys (https://openwrt.org/toh/linksys/e8450)
 
There are some known issues with whats here but Ill clean that up in follow up PRs, mainly:
  - Disable logd
  - Add back the backdoor network interface for the new mt-7622 hardware
  - disable conntrack and netfilter modules
  - Enable MIMO hardware offloading on belkin 3200
  - Supporting new `htmode` for wifi channel allocation (AX modes specifically)? however, after reviewing the docs we might just wanted to leave this in a general sweet spot: https://openwrt.org/docs/guide-user/network/wifi/basic#htmodewi-fi_channel_width
  - Enable radio3 for wifi6
> Each of these items has been spun out as an issue in our backlog

## Previous Behavior
- Only supported 2 APs architectures
- Weekly build of `ar71xx, ipq806x `
- Openwrt configs were built with the default being ar71xx

## New Behavior
- Support for 3 AP architectures: ar71xx, mt-7622, ipq806x (technically 4 with x86_64 as well but we dont run that on hardware)
- Default openwrt config leverages neutral x86_64 architecture so no architecture/hardware specific packages are pulled in. We will now default to leveraging x64_64 when building `common.config` for openwrt.
- Leveraging architecture specific `files-<arch>` dir to overlay config after `files` templating. This provides a much clearer way to place architecture specific configuration instead of `if, elif, ...` in the go templates
- DSA configuration for `mt-7622`
- Weekly build of: `ar71xx, mt-7622`
- Move `ar71xx` specific configuration (wired, wireless) to its own files-ar71xx folder
- Disable golden tests for `ipq806x` since we will not be investing any additional effort into that architecture

## Tests
- DSA network configuration was tested on the mgmt interface for both vlan 103 & 503

- brctl show, confirming STP enabled on all interfaces:
```
# brctl show
bridge name     bridge id               STP enabled     interfaces
scaleslow-br            7fff.e89f80e7f941       yes             br-lan.500
                                                        br-lan.100
br-lan          7fff.e89f80e7f941       yes             lan4
                                                        lan2
                                                        wan
                                                        lan3
                                                        lan1
mgmt-br         7fff.e89f80e7f941       yes             br-lan.503
                                                        br-lan.103
scalefast-br            7fff.e89f80e7f941       yes             br-lan.501
                                                        br-lan.101
```
> note spanning tree (STP) is enabled

Tested the uptime with real traffic for the last ~4 days. Looks stable:

```
# uptime
 08:39:56 up 4 days,  3:49,  load average: 0.14, 0.04, 0.01
```

CI image build results: https://github.com/socallinuxexpo/scale-network/actions/runs/5946721473


openwrt/configs/common.config
```diff
@@ -18,89 +18,53 @@ CONFIG_LUA_ECO_OPENSSL=y
...
-# CONFIG_PACKAGE_logd is not set
 CONFIG_PACKAGE_logrotate=y
-CONFIG_PACKAGE_lua=y
 # CONFIG_PACKAGE_nftables-json is not set
-# CONFIG_PACKAGE_odhcpd-ipv6only is not set
 CONFIG_PACKAGE_openssh-client=y
 CONFIG_PACKAGE_openssh-client-utils=y
 CONFIG_PACKAGE_openssh-keygen=y
 CONFIG_PACKAGE_openssh-server=y
-CONFIG_PACKAGE_python3-base=y
```
sarcasticadmin, line 80: removing `python3` since we dont currently rely on it for anything

openwrt/configs/common.config
```diff
@@ -18,89 +18,53 @@ CONFIG_LUA_ECO_OPENSSL=y
...
 CONFIG_PACKAGE_bash=y
 # CONFIG_PACKAGE_dnsmasq is not set
 # CONFIG_PACKAGE_dropbear is not set
 # CONFIG_PACKAGE_firewall4 is not set
 CONFIG_PACKAGE_iperf3=y
+# CONFIG_PACKAGE_jansson is not set
 CONFIG_PACKAGE_jq=y
-CONFIG_PACKAGE_kmod-ipt-core=y
-CONFIG_PACKAGE_kmod-nf-ipt=y
-CONFIG_PACKAGE_kmod-usb-serial=y
```
sarcasticadmin, line 40: Some of these kernel modules were for serial connectivity, but arent used since we just use the serial pins on the board instead of a dedicated usb device

openwrt/configs/common.config
```diff
@@ -18,89 +18,53 @@ CONFIG_LUA_ECO_OPENSSL=y
...
 CONFIG_PACKAGE_libreadline=y
-CONFIG_PACKAGE_librrd1=y
-CONFIG_PACKAGE_librt=y
-CONFIG_PACKAGE_libubus-lua=y
-CONFIG_PACKAGE_libuci-lua=y
+# CONFIG_PACKAGE_libucode is not set
 CONFIG_PACKAGE_libudev-zero=y
-CONFIG_PACKAGE_libuuid=y
 CONFIG_PACKAGE_lldpd=y
-# CONFIG_PACKAGE_logd is not set
```
sarcasticadmin, line 71: will add this back in a follow up PR

openwrt/files-mt7622/etc/config/network.0
```diff
@@ -0,0 +1,111 @@
+config interface 'loopback'
```
sarcasticadmin, line 1: Im opting to embed the configs for the network and the wireless vs template them. I dont think contorting the existing config is worth it given how static this typically is. Open to ideas here

openwrt/configs/common.config
owendelong: Uh, shouldn't we be going the other way... Deprecating Python2 and migrating all our dependencies to Python3?

openwrt/configs/x86-generic.config
owendelong, line 1: Do we really want to build x86 targets? For what?

openwrt/files-mt7622/etc/config/network.0
owendelong: I really don't understand the need for a ula_prefix in OpenWRT. If we can avoid it, I'd rather do so. We have no use for ULA in our environment. OTOH, it seems that OpenWRT may perhaps break badly without one for reasons utterly passing understanding. I haven't had time to dig into this and/or rant about it upstream, but if you can look at that and provide some feedback, that would be appreciated.

owendelong (commented):
I've provided individual comments inline, but generally looks good. Excellent effort, Rob.

openwrt/files-mt7622/etc/config/network.0
sarcasticadmin: thanks for the catch @owendelong This was just a carrier over from the default auto gen config from openwrt. Removing it doesnt break anything from my testing.

It comes from the uci defaults: https://github.com/openwrt/openwrt/blob/master/package/base-files/files/etc/uci-defaults/12_network-generate-ula but why thats a default I dont know either.

openwrt/configs/common.config
sarcasticadmin: this only pertains to python3 on the openwrt image. I opted to remove it since we dont currently have any python scripts installed on the APs. So we wont have python3 or python2 on the APs so we save the build time and the space on the image

openwrt/configs/x86-generic.config
sarcasticadmin: Youre right and we wont be using the target on any deployment but I briefly touched on it in the PR description:

> Default openwrt config leverages neutral x86_64 architecture so no architecture/hardware specific packages are pulled in. We will now default to leveraging x64_64 when building common.config for openwrt.

In order to do multiple architectures in openwrt I have to start with a base config which needs an architecture. It autoselects one if I dont have any defined. `x86` is a neutral architecture so we dont get any specific architecture or model packages. Theres no other way to filter them out either. If anyone else has better ideas how to do this Im all ears.

So instead the idea is to do the following order of operations to generate our configurations across N architectures:
```
    1. Generate a common.config using the x86 architecture since it omits
       any specific architecture packages. All packages declared here should
       be hardware independent and common across any architecture in scales
       environment.
    2. For new architectures generate a $TARGET-generic.config without
       including any common config.
    3. Combine $TARGET-generic.config + common.config and tweak until it
       includes all desired configuration. Output this the
       $TARGET-generic.config. This will include all CONFIG_TARGET config and
       architecture specific packages
 ```
 
Previously we were diffing the configs based on the prefixes of the build option names. Anything with that included CONFIG_TARGET was target specific, while everything else was common. This no longer works since there are many packages that exist specifically for a given arch/model and are not prefixed with CONFIG_TARGET.

**Comment by sarcasticadmin** [comment]
@owendelong thanks for the review and the feedback! If you need any more clarification you know where to find me!

File: openwrt/Makefile [hunk 5]
```diff
@@ -76,10 +78,13 @@ diffconfig:
...
 
 commonconfig: diffconfig
+	ifneq ($(TARGET), x86)
+	  $(error "common config can only be generated from x86")
+	endif
 	cat ./$(BUILD_DIR)/$(IMAGEBUILDER)/.diffconfig | grep -v CONFIG_TARGET > ./configs/common.config
 
 targetconfig: diffconfig
-	cat ./$(BUILD_DIR)/$(IMAGEBUILDER)/.diffconfig | grep CONFIG_TARGET > ./configs/$(TARGET)-generic.config
+	comm -23 <(sort ./$(BUILD_DIR)/$(IMAGEBUILDER)/.diffconfig) <(sort ./configs/common.config) > ./configs/$(TARGET)-generic.config
```
kylerisse, line 87: `comm`, nice!

openwrt/configs/common.config
kylerisse: makes sense

openwrt/configs/common.config
kylerisse: good riddance!

openwrt/configs/mt7622-generic.config
```diff
@@ -0,0 +1,25 @@
+# CONFIG_PACKAGE_kmod-ata-ahci-mtk is not set
```
kylerisse, line 1: Thanks for marking all the `not set` features. This really helps somebody unfamiliar with all the bells and whistles like me understand what's going on.

openwrt/docs/BUILD.md
```diff
@@ -212,3 +221,59 @@ x 2.6.32, with debug_info, not stripped
...
+```
+./scripts/diffconfig.sh | tee .diffconfig | grep -v CONFIG_TARGET > $(git root)/openwrt/configs/common.config
+```
+
+```
+comm -23 <(sort ./.diffconfig) <(sort $(git root)/openwrt/configs/common.config) > $(git root)/openwrt/configs/mt7622-generic.config
+```
+> Assuming mt7622 is our target arch
+
+Thats it now you should have a generic config and a arch specific config going forward. The majority of these steps are taken care of by the Makefile in the `openwrt` dir but in the cases where
```
kylerisse, line 278: nice, seems like a clean setup

facts/secrets/mt7622-openwrt-example.yaml
```diff
@@ -0,0 +1,21 @@
+# scale
+root_hash: "$1$zh0PjBbB$f9aFGDX9vNYNdSRexhib8/"
+
+# Bump for each year for scale
+scale: 19
```
kylerisse, line 5: 21?

kylerisse (approved):
@sarcasticadmin this all makes sense logically. As discussed, I didn't run through full test as I more than trust your methodology, and as you stated it's best to get this moving now since its getting beefy, and continue to iterate. Very cool idea to default to the common architecture and leverage arch specific configs. :ship:

facts/secrets/mt7622-openwrt-example.yaml
sarcasticadmin: Ya let me bump that to `21`, good catch!

sarcasticadmin:
Thanks @kylerisse for your review, Ill make the small tweaks and then report back when its push up

sarcasticadmin:
@owendelong @kylerisse Marking this as ready! Thanks again for the feedback and review. Ive spun out some additional items to address with this hardware after we get this initial set of changes merged.

facts/secrets/mt7622-openwrt-example.yaml
sarcasticadmin: Updated

owendelong (approved):
Looks good to me. Excellent work, Rob.

PR #619, opened 2023-09-21
Tweaked LibInventory.pm to produce human readable version of inventor…

owendelong:
Modified LibInventory.PM to produce human readable version of inventory.py output for testing/debugging purposes.

## Previous Behavior
silent non-functional library

## New Behavior
Remove main() to turn it into a functional library, but so far, not needed as a library.

## Tests
Ran it once, seemed OK.

### DIFF
--- a/ansible/LibInventory.pm
+++ b/ansible/LibInventory.pm
@@ -4,8 +4,9 @@
 #
 use strict;
 
-#use JSON;
+use JSON;
 use Data::Dumper;
+main();
 
 sub main
 {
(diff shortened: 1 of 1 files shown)


owendelong:
I think that the addition of the commit for this into #626 is accidental and that this PR is somewhat orthogonal to the intended content of that PR.

However, based on comments elsewhere from @kylerisse, it's entirely possible that this PR has been overtaken by events rendering it moot and we should (possibly) simply delete LibInventory.pm entirely.

sarcasticadmin:
Going to close this, I dont think we need this any longer.

PR #625, opened 2023-09-30
[READY] - nix: refactor and update

Kranzes:
## Description of PR

Relates to: https://github.com/socallinuxexpo/scale-network/issues/627

- Moving `perl` tests to be trigger via nix instead of from inside docker
- Refactoring nix to leverage [flake.parts](flake.parts) to utilize the module system for all aspects in the flake

### Update

Going to probably unify more of our gha tests into a single gha workflow.yml in a follow up.

## Previous Behavior
- Testing `perl` ci via docker
- Not using flake.parts within the flake

## New Behavior
- `perl` runs via nix checks in ci
- flake.parts and refactoring of the nix code for the flake

## Tests

- CI tests pass

Can run all of our tests in one go via:

```
nix flake check
```

PR #626, opened 2023-10-03
[READY] Od pre21x 01 -- Improvements to switch management system

owendelong:
## Description of PR
Major improvements and rewrites of the scripts to push configurations on to switches

## Previous Behavior
Kind of threw the config at the switch and hoped it would stick.

## New Behavior
Uses Expect and much better error checking. A single consolidated newly created script replaces all of the previous scripts for installing configurations on switches using different methods.

New script supports bulk/local, update in situ, and serial. Same script can install miniconfig over serial and then bulk load with different parameters.

## Tests
Code is still under development. Once it is complete, will make sure it lints and runs successfully. Hoping to be able to test at work party.


ansible/LibInventory.pm
```diff
@@ -4,8 +4,9 @@
 #
 use strict;
 
-#use JSON;
+use JSON;
```
sarcasticadmin, line 7: Does this PR supersede https://github.com/socallinuxexpo/scale-network/pull/619?

switch-configuration/config/scripts/switch_config_loader
```diff
@@ -0,0 +1,229 @@
...
+#        using "load override" functionality.
+#
+# update_switches
+#        A predecessor to override switches rendered obsolete by the development
+#        of override switches, long overdue for retirement.
+#
+#
+# New finctionality:
+#
+# command line usage:
```
sarcasticadmin, line 30: Would it be possible to add this comment section as a flag: `-h` for the switch_config_loader? I think that would be really helpful for anyone whose not super familiar with the tool.

sarcasticadmin (approved):
@owendelong thanks for consolidating the duplicated logic across the bulk, override, and update switch scripts.

Im excited to have a more robust feedback loop on our switch flashing process. Small suggestion around including a `-h` flag.

My only other concern is needing to include the new dependencies to make sure the CI has them and checks pass. Since with our previous CI build and/or the changes to the tests from #625 well need to include the new deps.  Happy to help on this front.

Overall, this PR looks great! Thanks for continuing to iterate on the switch config process

switch-configuration/config/scripts/switch_config_loader
owendelong: Yeah, I can add usage() and -h capabilities, that should be pretty easy.

ansible/LibInventory.pm
owendelong: I don't think so... I think it's orthogonal.

sarcasticadmin:
I just noticed this but I think we need to rebase this branch. It looks a little funky with the existing merge commit in this branch: 
ss-202310111696997817

Happy to help fix it, not rush just would like to do it prior to merging.

ansible/LibInventory.pm
owendelong: It's not unlikely, however, that the changes in 619 got included here due to my poor management of branches.

switch-configuration/config/scripts/switch_config_loader
owendelong: This has been done and pushed.

Issue #641, opened 2023-11-21
Openwrt seems to have a prometheus exporter.


nixinator:
Seems that openwrt has a prometheus exporter now.

Shall we add it , and then setup graphana and Prometheus to monitor and display.

could be cool.

don't know if the juniper gear has a promethus exporter.. be nice to enable if we do ?

https://grafana.com/blog/2021/02/09/how-i-monitor-my-openwrt-router-with-grafana-cloud-and-prometheus/

owendelong:
If someone is so inclined, they could experiment with this:
￼


I don’t know anything about it (or prometheus for that matter).

Owen

nixinator:
we just need to add this code to a semi powerful machine, as things get a quite busy on the monitoring side, and can create considerable cpu/disk load, however network is fairly ok.

```

{ config, pkgs, ... }: {
  # grafana configuration
  services.grafana = {
    enable = true;
    domain = "grafana.pele";
    port = 2342;
    addr = "127.0.0.1";
  };

  # nginx reverse proxy
  services.nginx.virtualHosts.${config.services.grafana.domain} = {
    locations."/" = {
        proxyPass = "http://127.0.0.1:${toString config.services.grafana.port}";
        proxyWebsockets = true;
    };
  };
}
```

on the openwrt side, i presume the Prometheus exporter is just an APK.


```
# hosts/chrysalis/configuration.nix
  services.prometheus = {
    enable = true;
    port = 9001;
  };
```

i don't have any openwrt boxes available, but if we can deploy the agent to a test box , we can take a look at the openwrt promethious agent, and see if it thrashes it, or it can run without any performance problems.

@rob , do you wanna have a crack at this on tmate sometime, if you have a hardware openwrt available to you?

@kylerisse , i think there might be a hookable api for events..?????

davidelang:
how does the data that the prometheus exporter provide compare with what we have 
configured zabbix to gather in the past from the openwrt systems?

I'd have to dig into the configs for specifics, but besides the standard linux 
stats it also include a lot of wifi related stats.

and then we also gather wifi details via syslog (dumps of what endpoints are 
connected that I remember, but I think there are a couple of other things as 
well)

David Lang

owendelong:
All else being equal, I’d generally rather use Apache than NGINX unless there’s some significant reason to do otherwise. Most of my interactions with NGINX have disabused me of my prior belief that Apache was the most difficult thing one could possibly try to configure.

Owen

[...]

Issue #567, opened 2023-03-10
monitoring and make nice cute graphs of network ....stuff


nixinator:
## Description

Monitoring of various network stats need to be configured and deployed.

whats my options.

## Acceptance Criteria

Nice graphs!!!

nixinator:
so doing some more research into this it seems

collectd running on the openWRT AP's feeding back to influx and graphana might be a good option.

It's quite declarative , and the configuration can be built from the single source from the scale-network repo.


https://blog.christophersmart.com/2019/09/09/monitoring-openwrt-with-collectd-influxdb-and-grafana/

We can probably monitor a whole  host of other things using this method.

I'll have a play and see what it looks like.

sarcasticadmin:
> collectd running on the openWRT AP's feeding back to influx and graphana might be a good option.
>
> It's quite declarative , and the configuration can be built from the single source from the scale-network repo.

I saw this in the past and it looks like a promising solution, especially if the entire config is declarative :+1: 

> I'll have a play and see what it looks like.

Let me know how it goes :+1:

sarcasticadmin:
Given that were utilizing prometheus a good dashboard starting place might be: https://grafana.com/grafana/dashboards/11147-openwrt/

sarcasticadmin:
Closing this as the initial work for getting grafana up and running has been completed. Any discuss around specific graphs should be discussed in #756

PR #642, opened 2023-11-25
[READY] - monitoring: grafana and prometheus service enabled

sarcasticadmin:
## Description of PR

Relates to: #641 #567 

This enables the foundations for grafana and prometheus services on the monitoring vm. I originally wanted to leverage `collectd` instead of prometheus but that ended up being more trouble than it was worth with influxdbv2 or graphite. Landing on prometheus should give us a good base and we can restrict the access to the webserver for scaping in a few ways. Grafanas admin password is currently being set on first login when creating a VM.

Currently we still need to:

  - Enable TLS
  - Generate a static configs for scapers
  - Enable prometheus exporters on Servers
  - Enable prometheus exporters on APs
  - Enable prometheus exporters on Pis

This will be done in follow up PRs if this initial approach is approved.

Prometheus is also added as a common nixos module so that it can be consumed from other machines.

ss-202311251700943674

## Previous Behavior
- Monitoring server was missing from scale-network automation

## New Behavior
- Monitoring server provisioned via a nixos config

## Tests

From another nixos machine:
```
nix build .#nixosConfigurations.monitor.config.system.build.vm -L
export QEMU_NET_OPTS="hostfwd=tcp::2222-:22"
./result/bin/run-nixos-vm
```

Setting up port forwarding to expose on `localhost:8000`:
```
ssh -o StrictHostKeyChecking=no -p 2222 rherna@127.0.0.1 -L 8000:127.0.0.1:80
```


nixinator:
Nice work on this Rob, i'm currently testing your changes... 


ON TLS 

Were going to require certs and i presume ACME is out because it's on internal ipv4 subnet.

Hmmm....

We could get everything running on ipV6, then we could get ACME , if acme supports  ipv6 however. This would require our internal servers to have public ipV6 DNS entries.. I don't know how i feel about that... maybe i should feel good.

Failing that were going to need our own CA, and start shunting certs about, which might be a massive PITA. However it would be a good opportunity to build a nix internal ACME system along the lines of  https://smallstep.com/blog/private-acme-server/

however, this may not be an insignificant amount of work.

davidelang:
perfect is the enemy of good enough :-)

David Lang

owendelong:
If people don’t mind importing my Root certificate (available at http://www.delong.com <https://www.delong.com/>), then I have a CA already set up that we can use to issue all the certs we need.

Owen

nixinator:
I have to remember that @owendelong doesn't use the internet, he *is* the internet. ;-)

owendelong:
Ah, but what will really bake your mind is that there is no such thing as “The Internet”.

What we call “The Internet” is a very loosely coupled set of networks that happen to have agreed to use the same protocol(s) and exchange traffic with each other.

Owen

owendelong:
> On Nov 27, 2023, at 15:58, Robert James Hernandez ***@***.***> wrote:
> 
> 
> Nice work on this @sarcasticadmin <https://github.com/sarcasticadmin> , i'm currently testing your changes...
> 
> @nixinator <https://github.com/nixinator> thanks for testing.
> 
> however, this may not be an insignificant amount of work.
> 
> At this point my plan is to do one of the following:
> 
> Issue a self signed certificate at runtime. This will require anyone connecting to web ui to "trust on first use" but thats good enough in my book for the small number of users for this service.
> Make grafana web service listen only on loopback and require that users to port forward via SSH to get to the web ui. Plus side here is that this doesnt require any self signed certs and we can leave it on http.
> Sounds like in general were good with this approach though, Ill mark it as READY
> 

Why not use my CA? Is there a problem with, perhaps using the same wildcard cert on every box?

Owen

sarcasticadmin:
> Why not use my CA? Is there a problem with, perhaps using the same wildcard cert on every box?

Its mainly to avoid the bootstrap problem of getting the private key on the VM in the first place. Many ways to do this but trying to keep it to a low level of effort.

owendelong (approved):
Most of this looks very "rusty". Ugh.

I guess we have to use the tools that are available.

owendelong:
> On Nov 27, 2023, at 23:18, Robert James Hernandez ***@***.***> wrote:
> 
> 
> Why not use my CA? Is there a problem with, perhaps using the same wildcard cert on every box?
> 
> Its mainly to avoid the bootstrap problem of getting the private key on the VM in the first place. Many ways to do this but trying to keep it to a low level of effort.
> 
TBH, I’m fine with just storing the private key in the server image. Since the cert isn’t being used for authentication and impersonation of our server is a pretty low risk scenario in this case, I don’t see guarding that private key as paramount.

owendelong:
I suspect we will eventually build a full time password vault of some form (IIRC, Rob had some thoughts in this area) and we can probably just use that.

Owen

nixinator:
sure my current flavour of secret management is secrix.

It's rather nice.. .

https://journal.platonic.systems/introducing-secrix/

PR #664, opened 2024-01-30
[READY] - Create gene_id_ed25519.pub

owendelong:
Gene Liverman public key

## Description of PR
Adding Gene's public key to facts.

## Previous Behavior
Gene couldn't log in

## New Behavior
Gene might be able to log in.

## Tests
Who tests things like this?


sarcasticadmin (approved):
Thanks for getting this add @owendelong :shipit:

PR #663, opened 2024-01-29
[READY] - Removing Ansible code but keeping tests_inventory.py

sarcasticadmin:
## Description of PR
The ansible code was last used during Scale 19x. Its remained in the codebase for reference and still had some usable tests for inventory.py. I think this is a good time to clean up this part of the codebase to make it clearer to the rest of the team what parts of the repository are actually being utilized.

## Previous Behavior
- `ansible` dir in the repo
- test_inventory.py was not being utilized by existing CI
- Off by 1 calculation for dhcp4range of /21

## New Behavior
- `ansible` dir removed and removed most references to ansible in the repo
- `test_inventory.py` moved over to `facts` and incorporated into existing CI
- All tests in `test_inventory.py` have been updated for latest functionality of `inventory.py`
- Off by 1 calculation fixed for dhcp4range

## Tests
- All tests from `test_inventory` CI tests are passing

Issue #592, opened 2023-03-15
define ntpd daemon inside the core role


sarcasticadmin:
## Description

Previously in the ansible code base we had defined ntpd: https://github.com/socallinuxexpo/scale-network/blob/ffda5117e0fd7757dbf4d43f190f13f5092b0254/ansible/roles/chrony/tasks/main.yml

We should define this again for the `core` role so that we can offer it as a service to the rest of the fleet (servers, aps, switches, etc.)

## Acceptance Criteria
- `core` role has ntpd service accessible
- nix tests added to confirm ntpd is working as expected

nixinator:
might be nice to enable authentication on this.
but it means loading a key onto the ntp clients and the servers. which is PITA. A age encrypted secret might suffice.

i can confirm that chronyd is in nixpkgs.

https://github.com/NixOS/nixpkgs/blob/nixos-22.11/nixos/modules/services/networking/ntp/chrony.nix

davidelang:
I've had bad experience with chronyd, please no.

David Lang

nixinator:
Tell us more! What bad experience did you have...

davidelang:
This was a couple jobs ago, and I don't remember all the details (I was a 
dedicated splunk admin there), but it was a combination of not generating useful 
logs for troubleshooting issues, and not reliably setting the time correctly 
(causing us to need to try and troubleshoot things)

David Lang

nixinator:
We i will generate a couple of configurations for different time servers..

let the best time server win! I've seen time server go badly wrong when running under some kinds of virtualisation.. it as unexpected side effects, but that was many moons ago. Software always gets better, where hardware fail eventually.

 Are you sure you were not under a MITM or a WITM attack? People love to do that !

davidelang:
very sure, we had an on-prem GPS server we were referencing against.

David Lang

nixinator:
@owendelong , if you want to have experiment with https://en.wikipedia.org/wiki/Precision_Time_Protocol .

However it looks pretty interesting, maybe a bit overkill . It also needs hardware support, can the Juniper platform do it?

I also went down a time sync research rabbit hole, I guess a time rabbit hole, which could be described as black hole, which was a bit of a 'time sink', rather than 'time sync'.

I found this is being used for synchronising clocks at layer one by encoding the time into the layer 1 encoding! Blimey...

funny enough it's called the white rabbit network. This is probably useful it we plan to run a nuclear fusion reactor at scale40x


[...]

PR #665, opened 2024-01-30
[READY] - enable ntpd on core machines

sarcasticadmin:
## Description of PR

Related to: #592

Getting ntpd enabled and running on the core machines

## Previous Behavior
- ntpd was not running on the core machines
- left old ansible reference after #663 in scaleInventory pkg

## New Behavior
- ntpd running on core and restricted to scale ipv4 and ipv6 ranges
- cleaned up old ansible ref  in scaleInventory pkg

## Tests
- Running vm locally confirms that is working as expected:

```
nix build .#nixosConfigurations.coreMaster.config.system.build.vm
```

- Running the vmtest for core works as expect:

```
$ nix build -L .#checks.x86_64-linux.core
```

### DIFF
--- a/nix/machines/core/common.nix
+++ b/nix/machines/core/common.nix
@@ -13,7 +13,7 @@
     useDHCP = false;
     useNetworkd = true;
     firewall.allowedTCPPorts = [ 53 67 68 ];
-    firewall.allowedUDPPorts = [ 53 67 68 ];
+    firewall.allowedUDPPorts = [ 53 67 68 123 ];
   };
 
   security.sudo.wheelNeedsPassword = false;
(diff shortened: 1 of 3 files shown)


owendelong:
Utterly unnecessary in our environment. For what we are doing, NTP is more than sufficient.Geeking out with PTP might be fun in another context, but here it would just be a distraction. OwenOn Feb 1, 2024, at 06:45, Lee Hughes ***@***.***> wrote:﻿
I've always want to try out ptp. Might be fun.
https://linuxptp.sourceforge.net/

nix/machines/core/common.nix
owendelong, line 1: That should probably be restrict 2001:470:f026::/48 rather than /64. If we limit to /64, we will block all of the other show subnets.

owendelong (changes requested):
See note on common.nix

nix/machines/core/common.nix
sarcasticadmin: Good catch @owendelong, PR has now been updated

owendelong (approved):
Thanks for getting this done, Rob.

Issue #660, opened 2024-01-25
qemu/kvm nixos configuration to replace bhyve hypervisor


sarcasticadmin:
## Description

we'll be moving the hypervisor to kvm in an effort to unify the configuration of the hosts and the vms.

Most of this work was done at the work party 1/13/24-1/14/24

## Acceptance Criteria
- working nixos kvm host configuration for hypervisor physical hosts (non nixos VMs)
- microvm (utilizing qemu) for nixos VMs

sarcasticadmin:
Closing as this was completed for 21x

PR #666, opened 2024-01-30
[READY] - init devServer 

sarcasticadmin:
## Description of PR
Relates to: #660

init the configuration of the `devServer` thats being hosted in northern california

## Previous Behavior
- No dev server :/

## New Behavior
- dev server for all kinds of fun
- base.nix for setting the bulk of base packages on our systems

## Tests
- dev server booted and can `nixos-rebuild switch`

### DIFF
--- a/nix/machines/devServer/default.nix
+++ b/nix/machines/devServer/default.nix
@@ -1,0 +1,39 @@
+{ config, pkgs, ... }:
+
+{
+  # remove the annoying experimental warnings
+  nix.extraOptions = ''
+    experimental-features = nix-command flakes
+  '';
+
+  networking = {
+    useNetworkd = true;
+    useDHCP = false;
+    firewall.enable = true;
+  };
+
+  systemd.network = {
+    enable = true;
+    networks = {
+      "10-lan" = {
+        matchConfig.Name = "eno1";
+        enable = true;
+        networkConfig.DHCP = "yes";
+      };
+    };
+  };
+
+  environment.systemPackages = with pkgs; [
+    wget
+    git
+    vim
+    efibootmgr
+    gptfdisk
+    screen
+  ];
+
+  services.openssh = {
+    enable = true;
+    openFirewall = true;
+  };
+}
--- a/flake.lock
+++ b/flake.lock
@@ -20,6 +20,46 @@
         "type": "github"
       }
     },
+    "flake-utils": {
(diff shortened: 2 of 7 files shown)


nix/machines/devServer/default.nix
owendelong, line 1: Given that this is the hypervisor and KEA runs as a VM under the hypervisor, it might be best to use a static address for eno1 on this host.

IT appears to me that the tcpdump and mtr packages are not listed, is this an oversight or did I miss something?
Key tools that really should work on all of our hosts include:

- mtr
- tcpdump
- dig
- host
- arp
- ip (iproute2 package usually)
- netstat
- fuser
- lsof
- ssh (client)
- ssh (server) in most if not all cases
- Probably others I am forgetting at the moment, but those are the ones that have hit me hardest on nix systems so far.

owendelong (changes requested):
Other than as mentioned above, looks good to me.

nix/machines/devServer/default.nix
sarcasticadmin: This PR was originally reflecting the current state of whats been applied to the machine.

> Given that this is the hypervisor and KEA runs as a VM under the hypervisor, it might be best to use a static address for eno1 on this host.

Sure what address would you like to use?

> IT appears to me that the tcpdump and mtr packages are not listed, is this an oversight or did I miss something?
Key tools that really should work on all of our hosts include:

We can add all the packages we want, this is a good list to start. Ill add the additional configuration to this PR for these packages. Packages like ssh-server are already taken care of when enabling the ssh server.

owendelong: Awesome. I think the same addresses we previously assigned to the BHyve hypervisors should be good, right?

sarcasticadmin: Cleared up the confusion in a sidebar conversation: we'll be leaving `eno1` as dhcp since this is only for the dev Server and not for the hypervisors. That will be in a follow up PR.

owendelong: So should I approve and merge this as is or wait for you to mark it ready and re-review?

sarcasticadmin: just pushed up the common packages you mentioned, so this is ready for another review

owendelong (approved):
Looks good to me.

PR #661, opened 2024-01-28
[Ready] Added last-minut request for Expo Hal VM Vending Machine (against my better judgment)

owendelong:
## Description of PR
Add configuration to support VM Vending Machine in Expo Hall

## Previous Behavior
We didn't have to worry about DefCon bringing complexity to our network.

## New Behavior
Should (hopefully) support last minute request for a vending machine to support short term VMs for people to use for mischief.

## Tests
There's no way to test this in advance, we just have to cope with whatever happens as it comes up.

PR #667, opened 2024-02-02
[READY] 21x graphics 01

owendelong:
## Description of PR
Adding some of the graphics for 21x

## Previous Behavior
Pre-21x graphics only

## New Behavior
Some graphics for 21x

## Tests
It looks sort of OK in Inkscape.


sarcasticadmin (changes requested):
@owendelong can we remove the `.png` files? Id rather not commit binaries to the repo if we can avoid it.

Also the commit history looks odd, can you run `git checkout master && git pull --r origin master && git checkout 21x_graphics_01 && git rebase master` . The working branch should never have `master` merged into it.

owendelong:
I'd rather not have binaries, too, but unfortunately, the artist insists on workin in PNG, so these are the actual originals and the SVGs are somewhat lower fidelity replicants I made from the PNGs. As such, I'd rather not lose the originals.

I'll try to hork the history as requested.

sarcasticadmin:
You were on some intermittent branch from on #661 . Not sure how that all happened but easy enough to fix.

So what I did:

```
$ git fetch upstream/21x_graphics_01 # get a copy of the remote branch
$ git checkout 21x_graphics_01 # checkout that remote branch
$ git reset --hard master # reset hard to local master since git rebase master failed with lots of old conflicts
$ git cherry-pick c33b898319701a12fa704589fb2d53560cd71d07 # The commit where you added the graphics thats still in the local remote branch upstream/21x_graphics_01
$ git push -f upstream 21x_graphics # Make remote history match our local one
```

sarcasticadmin (approved):
Not too excited about added binaries to the repo, but Im ok with it as long as its kept to a minimum. I wouldnt want this repo to become cumbersome to clone.

PR #669, opened 2024-02-13
[READY] [Issue-567] - Adding OpenWRT Grafana Dashboard

### LINKED ISSUE #567: monitoring and make nice cute graphs of network ....stuff

jshcmpbll:
## Description of PR
relates to: #567 

## Previous Behavior
No dashboards configured by default for grafana

## New Behavior
openwrt grafana dashboard

## Tests
nix build
@sarcasticadmin is going to test with a local OpenWRT deployment since I dont have any hardware

### DIFF
--- a/nix/machines/monitor.nix
+++ b/nix/machines/monitor.nix
@@ -1,6 +1,7 @@
 { config, lib, pkgs, inputs, ... }:
 let
   hostname = "monitoring.scale.lan";
+  dashboard = pkgs.copyPathToStore ../../monitoring/openwrt_dashboard.json;
 in
 {
   imports =
@@ -69,6 +70,12 @@ in
             type = "prometheus";
             access = "proxy";
             url = "http://127.0.0.1:${toString config.services.prometheus.port}";
+          }
+        ];
+        dashboards.settings.providers = [
+          {
+            name = "openwrt";
+            options.path = dashboard;
           }
         ];
       };
--- a/nix/machines/monitor.nix
+++ b/nix/machines/monitor.nix
@@ -65,6 +65,7 @@ in
         datasources.settings.datasources = [
           {
             name = "prometheus";
+            uid  = "P1809F7CD0C75ACF3";
             type = "prometheus";
             access = "proxy";
             url = "http://127.0.0.1:${toString config.services.prometheus.port}";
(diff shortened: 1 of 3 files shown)


nix/machines/monitor.nix
```diff
@@ -65,11 +66,18 @@ in
         datasources.settings.datasources = [
           {
             name = "prometheus";
+            uid  = "P1809F7CD0C75ACF3";
```
kylerisse, line 69: just to clarify, does this value have any specific meaning or does it just need to be $SOME_VALUE that matches all the datasource references form the dashboard json?

sarcasticadmin: Im unclear on this myself

@jshcmpbll might be able to add some insight

jshcmpbll: @kylerisse Yup! The value itself has no real meaning but would need to match the data source values found in the dashboard like so:

https://github.com/socallinuxexpo/scale-network/blob/9258aee2c9590d4bfe924cad478c6c2b6debbe05/monitoring/openwrt_dashboard.json#L72

Each time I started the VM it seemed to use that UID for the datasource by default but I figured we might as well set it so there isn't a possibility of some change in the future resulting in the UID changing.

jshcmpbll:
> @jshcmpbll This looks like exactly what we need!
> 
> Going to work on getting this connected to openwrt shortly

I started working on the exporter replay test we talked about but still trying to figure out networking for nix tests.

owendelong (approved):
LGTM

owendelong:
Not merging because marked "REVIEW". If you want it merged, please mark it READY.

nix/machines/monitor.nix
kylerisse: sweet, thanks @jshcmpbll

sarcasticadmin (approved):
I was able to setup an AP and run this monitoring vm. The dashboard populates as expected and everything seems to be coming in nicely! (ignore the wifi quality metric, I really didnt gather too much data for this initial test)

ss-202402231708670002

I added a small change in the inventory.py to ensure that we are using `port 9100` for the metrics exporters on the APs.

Nice work @jshcmpbll ! This is a big step for getting the dashboard built out!

Issue #596, opened 2023-03-23
enable dhcpv6 server for kea on core machine


sarcasticadmin:
## Description

Relates to: https://github.com/socallinuxexpo/scale-network/issues/392

Kea config is currently only v4. Need to add our v6 version of the config.

## Acceptance Criteria
- Kea service running on core provides dhcpv6 server
- Appropriate vm test to match

nixinator:
Just our of curiosity, i thought V6 address were big enough so you don't need dhcp at all, and one less service is one less thing to go wrong?

davidelang:
we are passing some parameters via dhcp, so those are still needed.

David Lang

sarcasticadmin:
@nixinator Yes @davidelang is correct. Ultimately well we passing the v4 options via v6.

IIRC @owendelong had some other use cases for this as well

owendelong:
Yes... Sometimes you want a particular host to consistently get the same address. While EUI-64 sort of allows this, it's no longer default on the vast majority of systems. Using DHCP6 reservations provides a consistent way to manage these addresses.

PR #252, opened 2019-05-25
[READY] Resolve some open Ansible issues

kylerisse:
## Description of PR
Fixes #207
Fixes #226
Fixes #232

This PR includes fixes for 3 open issues as well as some general housekeeping within the Ansible directory.

## Previous Behavior
- isc-dhcp-server6 service would go OOM during startup or shortly after due to unnecessarily large IPv6 scope space
- IPv4 leases could be renewed for up to 15 minutes on a 5 minute lease, preventing APs from getting timely DHCP related configuration option updates
- Data files were tab separated instead of comma 
- Python2 was still being installed on servers
- Ansible file layout was unnecessarily cumbersome

## New Behavior
- Greatly reduce the IPv6 DHCP scope sizes to address issue #207 . DHCP scope is now 1024 per subnet for each DHCP server. The service no longer crashes with coredump/OOM conditions.
- Set the IPv4 DHCP max lease time equal to the default lease to address issue #226 . The server has no control over clients requesting renewals within the 5 minutes but will no longer allow them after a full first lease.
- Switched fact files from TSV to CSV to address issue #232 . This includes the lists for servers, routers, aps, and pis.
- Removed the python2 role from the project. It will be deprecated before the next SCaLE.
- Simplified the ansible file layout by removing the embedded etc/ansible directory.
- Added a header row to the AP list

## Tests
Test for Issue #207 
1. bring environment up with vagrant up
2. login to server1 with vagrant ssh server1, elevate to root
3. open file to edit /etc/netplan/50-vagrant.yml
4. add item to addresses list - 2001:470:f325:503::5/64 and save file
5. netplan apply
6. verify 2001:470:f325:503::5/64 attached to enp0s8
7. restart isc-dhcp-server6 with systemctl restart isc-dhcp-server6
8. tail dhcp log with tail -f /var/log/dhcp.log
9. in another terminal log in to server8 with vagrant ssh server8, elevate to root
10. created 200 test dhcp requests, for i in {1..200}; do dhclient -6; done;
11. in server1 terminal verify logs
12. verify isc-dhcp-server6 has not crashed with systemctl status isc-dhcp-server6 and viewing the "Active" item

Test for Issue #226 
1. login to server1 with vagrant ssh server1
2. review /etc/dhcp/dhcpd.conf and note that default-lease-time and max-lease-time now match

Test for Issue #232 
[...]

PR #671, opened 2024-02-19
[READY] - kea dhcpv6 config and service

sarcasticadmin:
## Description of PR

Fixes: #596
Depends on: #670 (Will have to rebase after)

Kea DHCPv6 using all of the existing vlans from switch configurations

A lot of this was made very easy by previous work (thanks @kylerisse ): https://github.com/socallinuxexpo/scale-network/pull/252 ,inventory.py ipv6 support, and the retired isc-dhcpv6. This was all very useful as reference for this implementation.

## Previous Behavior
- kea didnt have dhcpv6
- lease time was set to 4000 sec for dhcpv4

## New Behavior
- kea dhcpv6 configuration ([dhcp6-server.conf.txt](https://github.com/socallinuxexpo/scale-network/files/14326060/dhcp6-server.conf.txt) (ignore the .txt apartly github doesnt let you upload a .conf :laughing:)
- kea dhcpv6 service enabled on `core`
- dhcpv6 specific runNixOSTest for core test
- kea dhcpv4/v6 lease time for mgmt vlan is now 5 min
- kea dhcpv4/v6 lease time for all other subnets is 24min
- updates to python facts tests for new dhcpv6 configurations

## Tests
- Core sanity check in runNixOSTests runs well:

```
$ build -L .#checks.x86_64-linux.core
```

### DIFF
--- a/facts/inventory.py
+++ b/facts/inventory.py
@@ -168,7 +168,7 @@ def ip6toptr(ip_address):
 def dhcp6ranges(prefix, bitmask):
     """generates start and end IPv6 addresses for DHCP ranges"""
     if bitmask == 0:
-        return ["", "", "", ""]
+        return ["", ""]
     prefsplit = re.split(r"\:\:", prefix)[0]
     return [
         prefsplit + ":d8c::1",
--- a/facts/inventory.py
+++ b/facts/inventory.py
@@ -534,21 +534,29 @@ def generatekeaconfig(servers, aps, vlans, outputdir):
         for ap in aps
     ]
 
-    subnets_dict = [
-        {
-            "subnet": vlan["ipv4prefix"] + "/" + str(vlan["ipv4bitmask"]),
-            # generating uniq id (prefix with dots) for each subnet block to ensure autoids dont effect reordering
-            # called out in: https://kea.readthedocs.io/en/latest/arm/dhcp4-srv.html#ipv4-subnet-identifier
-            # subnet ids must be greater than zero and less than 4294967295
-            "id": int(vlan["ipv4prefix"].replace('.', '')),
-            "user-context": { "vlan": vlan["name"] },
-            "pools": [{"pool": vlan["ipv4dhcpStart"] + " - " + vlan["ipv4dhcpEnd"]}],
-            "option-data": [{ "name": "routers", "data": str(vlan["ipv4router"]) }],
-        }
-        for vlan in vlans
-        if vlan["ipv4bitmask"]
-        != "0"  # Make sure to skip vlans that have no ranges
-    ]
+    subnets_dict = []
+    for vlan in vlans:
+        # Make sure to skip vlans that have no ranges
+        if vlan["ipv4bitmask"] == "0":
+            continue
+        else:
+            subnet = {
+                "subnet": vlan["ipv4prefix"] + "/" + str(vlan["ipv4bitmask"]),
+                # generating uniq id (prefix with dots) for each subnet block to ensure autoids dont effect reordering
+                # called out in: https://kea.readthedocs.io/en/latest/arm/dhcp4-srv.html#ipv4-subnet-identifier
+                # subnet ids must be greater than zero and less than 4294967295
+                "id": int(vlan["ipv4prefix"].replace('.', '')),
+                "user-context": { "vlan": vlan["name"] },
+                "pools": [{"pool": vlan["ipv4dhcpStart"] + " - " + vlan["ipv4dhcpEnd"]}],
+                "option-data": [{ "name": "routers", "data": str(vlan["ipv4router"]) }],
+            }
+            # lower lease times for the APs
+            if vlan["name"] in [ "cfInfra", "exInfra"]:
+                # Set lifetime of lease to always be 300 seconds
+                subnet["valid-lifetime"] = 300
+                subnet["min-valid-lifetime"] = 300
+                subnet["max-valid-lifetime"] = 300
+            subnets_dict.append(subnet)
 
     kea_config["Dhcp4"]["subnet4"] = subnets_dict
     kea_config["Dhcp4"]["reservations"] = reservations_dict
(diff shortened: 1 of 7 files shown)


facts/inventory.py
```diff
@@ -73,7 +73,7 @@ def genvlans(line, building):
     elems = re.split(r"\t+", line)
     rangeb, rangee = elems[2].split("-")
     for i in range(int(rangeb), int(rangee) + 1):
-        ipv6prefix = elems[3].split("/")[0].split(rangeb + "::")[0] + str(i) + "::"
```
sarcasticadmin, line 76: This was to match the changes in VVRNG from https://github.com/socallinuxexpo/scale-network/pull/661/files#diff-077a6c63504cfb184eac78e4bd46b53e79390027b942f021589bfda90a79e1abR21

facts/inventory.py
```diff
@@ -166,13 +166,21 @@ def ip6toptr(ip_address):
...
 def dhcp6ranges(prefix, bitmask):
-    """generates start and end IPv6 addresses for 2x DHCP ranges"""
+    """generates start and end IPv6 addresses for DHCP ranges"""
     if bitmask == 0:
-        return ["", "", "", ""]
+        return ["", ""]
     prefsplit = re.split(r"\:\:", prefix)[0]
+    # TODO: Infer this going forward
+    # Conditional for the exVmVendor vlan, matching the ipv4 /20
+    if prefix == "2001:470:f026:112::":
```
kylerisse, line 175: agree with TODO, but good iterative compromise for now.

facts/test_inventory.py
```diff
@@ -160,7 +160,7 @@ def test_bitmasktonetmask():
 def test_genvlans():
     '''test cases for the genvlans() function'''
     cases = [
-        ["VVRNG\ttest_vlan_\t\t200-201\t2001:470:f325:200::/54\t10.2.0.0/15\tdynamic vlan", [
+        ["VVRNG\ttest_vlan_\t\t200-201\t2001:470:f325::/48\t10.2.0.0/15\tdynamic vlan", [
```
kylerisse, line 163: thanks for including and updating testing

kylerisse (approved):
Back to supporting DHCP6 and without all the hassle of isc-dhcpd6. Thanks @sarcasticadmin

facts/inventory.py
owendelong: Makes sense.

facts/inventory.py
owendelong: Agreed.

facts/inventory.py
```diff
@@ -509,9 +551,32 @@ def generatekeaconfig(servers, aps, vlans, outputdir):
...
         f.write(json.dumps(kea_config, indent=2))
 
+    subnets6_dict = [
+        {
+            "subnet": vlan["ipv6prefix"] + "/" + str(vlan["ipv6bitmask"]),
+            # generating uniq id (prefix with dots) for each subnet block to ensure autoids dont effect reordering
+            # called out in: https://kea.readthedocs.io/en/latest/arm/dhcp4-srv.html#ipv4-subnet-identifier
+            # subnet ids must be greater than zero and less than 4294967295
+            #
+            # for ipv6 well take the last 8 hex digits to make sure the id is small enough but uniq
```
owendelong, line 565: Given that we are starting from a /48 (and utterly unlikely to ever go beyond that) and issuing /64 VLANs, the last 4 hex digits is sufficient to be guaranteed unique.

Obviously, we need to special case the ...:0::/64 prefix if we ever use that for something DHCP, but otherwise, we should be able to use the straight subnet number.

facts/test_inventory.py
owendelong: I'm slightly concerned about the use of 2001:470:f325:: throughout this as the current prefix is 2001:470:f026::/48.

owendelong (changes requested):
See inline comments.

Also, would it be possible to arrange so that dynamically issued (not reservations) DHCPv6 addresses come from a pool in each subnet numbered 2001:470:f026:<subnet>:d8c:<x>:<y>:<z> where X/Y/Z are dynamically assigned?

facts/test_inventory.py
kylerisse: I believe the intent of this unit test is to validate *ANY* /48 and not necessarily the real one. @sarcasticadmin can probably clarify if the intent I read was different.

sarcasticadmin: @kylerisse correct, this test is a generic case and the prefix doesnt hurt the validity of the test. 

We can update it to avoid confusion and that probably would be a good first issue for someone :slightly_smiling_face: since I think there a quite a few references to the old prefix.

facts/inventory.py
sarcasticadmin: thats sounds good, well go with 4 hex digits

sarcasticadmin:
@owendelong thanks for the comments.

> Also, would it be possible to arrange so that dynamically issued (not reservations) DHCPv6 addresses come from a pool in each subnet numbered 2001:470:f026::d8c::: where X/Y/Z are dynamically assigned?

Yes I believe this is possible and we would just have to toggle the `"reservations-in-subnet": False` kea config option and/or the other related options.

So as an example: our dynamic range for `vlan 112` would be:
```
2001:470:f026:112:d8c::1
2001:470:f026:112:d8c::800
```
>`d8c` because its close to dhcp :wink: 

Then have a reservation within `vlan 112` would be:

```
2001:470:f026:112:::100
```

Just want to make sure Im following here.

owendelong:
> @owendelong thanks for the comments.
> 
> > Also, would it be possible to arrange so that dynamically issued (not reservations) DHCPv6 addresses come from a pool in each subnet numbered 2001:470:f026::d8c::: where X/Y/Z are dynamically assigned?
> 
> Yes I believe this is possible and we would just have to toggle the `"reservations-in-subnet": False` kea config option and/or the other related options.
> 
> So as an example: our dynamic range for `vlan 112` would be:
> 
> ```
> 2001:470:f026:112:d8c::1
> 2001:470:f026:112:d8c::800
> ```
> 
>  > `d8c` because its close to dhcp 😉
> 
> Then have a reservation within `vlan 112` would be:
> 
> ```
> 2001:470:f026:112:::100
> ```
> 
> Just want to make sure Im following here.

Yes, that's correct.

owendelong:
I don't care how wide we make the dynamic range within the VLAN (up to 48 bits), TBH. So ::1 ::800 is fine, but so would ::0 through :ffff:ffff:ffff as far as I'm concerned.

(The first would result in <prefix>:d8c::1 through <prefix>:d8c::800, the latter would result in <prefix>:d8c::1 through <prefix>:d8c:ffff:ffff:ffff). 

I'd also be fine if we started from ::0, since in IPv6, there's nothing special about the ::0 address.

sarcasticadmin:
> (The first would result in :d8c::1 through :d8c::800, the latter would result in :d8c::1 through :d8c:ffff:ffff:ffff).

My initial hesitation here was due to the problems we saw with `isc-dhcpv6` when the range was so large: https://github.com/socallinuxexpo/scale-network/pull/252 This is me being superstitious but we wont need all of those addresses anyway.

> I'd also be fine if we started from ::0, since in IPv6, there's nothing special about the ::0 address.

Good point

owendelong:
As I stated, I'm fine with the current smaller range and fine with it as is. Let me know when you've made the <prefix>:d8c: change and I'll approve and merge.

owendelong (approved):
Looks great, Rob. Thanks for addressing my nits.

facts/inventory.py
sarcasticadmin: This was updated to 4

sarcasticadmin:
@owendelong thanks for helping debug the kea dhcpv6 stuff the other day, saved me from banging my head against the wall.

Theres one small limitation in this config related to dhcpv6 relaying and Ill be spinning out for a follow up: Since we cant declare the same interface (eth0) in two subnet6 blocks well have to dynamically allocate the local subnet6 block and interface for the corresponding expo and conf core. The core will only ever be local to the subnet that its in the building for.

This is called out here in the kea config: https://kea.readthedocs.io/en/kea-2.2.0/arm/dhcp6-srv.html#ipv6-subnet-selection

For now I have the interface set to expo since thats how the core runNixOSTest is configured and theres good value were getting out of those test cases for confirming dhcpv6 is working.

Other than that I think this is ready to go @kylerisse @owendelong

Issue #630, opened 2023-10-14
Rename gha workflow to match existing functionality 


sarcasticadmin:
## Description

Currently we have validate-datafiles.yml: https://github.com/socallinuxexpo/scale-network/tree/4699e5e24d6dc3fe9dfc137c83218290ad84c322/.github/workflows but this contains not only the pytest for validation but additional tests to leverage the nix job cache. This results in our entire CI running in under 60 seconds for every change vs have to specify the paths like our previous CI workflows.

Given how expressive we can be via the tests inside the our nix configuration: https://github.com/socallinuxexpo/scale-network/blob/4699e5e24d6dc3fe9dfc137c83218290ad84c322/nix/tests/flake-module.nix I dont see much need for separate CI workflows and the typical features you would require for a workflow.

I would purpose we move rename `validate-datafiles.yml` to `ci.yml` or something generic and just continue to add out runs to the single job. 

## Acceptance Criteria
- ci.yml contains all ci jobs that we have

owendelong:
Seems reasonable to me... Please make it so and close issue.

PR #672, opened 2024-02-19
[READY] - rename gha CI to be generic

sarcasticadmin:
## Description of PR

Fixes: #630 

The existing CI workflow was originally for data validation but its been generic in nature for awhile

## Previous Behavior
- data validation was the name given the workflow

## New Behavior
- gha workflow now generically called `CI` to reflect its function

## Tests
- CI still works


kylerisse (approved):
Makes sense. Thanks @sarcasticadmin

PR #649, opened 2024-01-14
[READY] Generate prometheus config from inventory

jshcmpbll:
## Description of PR
Adding a new function to the inventory script to generate a Prometheus config for the monitor nix machine

## Previous Behavior
No Prometheus config generation

## New Behavior
Prometheus config generation

## Tests
nix build

PR #677, opened 2024-03-02
[READY] - splitting aplist into aps and apuse

sarcasticadmin:
## Description of PR

Supersedes: #585

Based on the conversations in https://github.com/socallinuxexpo/scale-network/pull/585 were opting to split up the aplist
separate `aps.csv` (serial, mac) and `apuse.csv` (room name, serial, address, wireless channel, etc.)

The idea being that we can make it easier and less error prone if we only have to update `apuse.csv`

## Previous Behavior
- aplist.csv was source of truth

## New Behavior
- `aps.csv` and `apuse.csv` are source of truth
- aps will be referable by their serial number as a `CNAME`

## Tests
- Tests pass in CI
- `nix flake check` passes


davidelang:
This is a good step from the single aplist, but 585 went a step further to split out room facts

1. what the IP range is for that building
2. what map that room is on
3. coordinates of where on that map the room is

the first is useful so that IP addresses can be auto-generated (even the two file split is less of a problem than the original, since the IP can be set to the AP name instead of to the serial/mac)

the other two are only going to be useful when we try to do a graphical monitoring display (and need to be combined with the coordinates within the room)

speaking of monitoring, with Zabbix, we had the APs auto-register themselves, is the new monitoring system getting pre-loaded with the devices to monitor or scanning the network to discover them?

davidelang:
Also, should we be creating aplist.csv? or should the things that currently use aplist.csv be modified to use the separate files?

sarcasticadmin:
@davidelang 

> 1. what the IP range is for that building

This is already identifiable in switch-configuration: https://github.com/socallinuxexpo/scale-network/blob/e857c20374465b3b1b810c70c8339a65286b4156/switch-configuration/config/vlans.d/Conference

> 2. what map that room is on
> 3. coordinates of where on that map the room is

These are data points that dont yet exist. And should be included in a new issue about AP roommap, deployment location, etc..

> the first is useful so that IP addresses can be auto-generated (even the two file split is less of a problem than the original, since the IP can be set to the AP name instead of to the serial/mac)

I dont think including the entire address is a problem. In the case of a failure, individuals wont have to modify the IP addresses during the show only update serials.

We could infer it from the switch-configurations but I think this becomes too obscure. Since now youre creating a composite unique identifier of the building + last octet.

> the other two are only going to be useful when we try to do a graphical monitoring display (and need to be combined with the coordinates within the room)

This should be a separate issue since the graphical display hasnt been built.

> speaking of monitoring, with Zabbix, we had the APs auto-register themselves, is the new monitoring system getting pre-loaded with the devices to monitor or scanning the network to discover them?

This is already taken care of by https://github.com/socallinuxexpo/scale-network/pull/649 which consumes the aps from pkgs.scaleInventory (inventory.py), builds an appropriate config and scrapes the AP prometheus endpoint.

> Also, should we be creating aplist.csv? or should the things that currently use aplist.csv be modified to use the separate files?

No, its no longer need. Im proposing with this PR to combine `apuse.csv` and `aps.csv` within `inventory.py`. As proposed in https://github.com/socallinuxexpo/scale-network/pull/585#issuecomment-1801099169 All of the necessary changes are within this PR. There isnt anything else that makes use of the `aplist.csv` that Im aware of.

davidelang:
On Sun, 3 Mar 2024, Robert James Hernandez wrote:

> @davidelang
>
>> 1. what the IP range is for that building
>
> This is already identifiable in switch-configuration: https://github.com/socallinuxexpo/scale-network/blob/e857c20374465b3b1b810c70c8339a65286b4156/switch-configuration/config/vlans.d/Conference

how do you know what room name is in which building? that's the roommap.csv file 
that was being used

>> 2. what map that room is on
>> 3. coordinates of where on that map the room is
>
> These are data points that dont yet exist. And should be included in a new issue about AP roommap, deployment location, etc..

the data doesn't exist, the file contains the placeholders for them.

>> the first is useful so that IP addresses can be auto-generated (even the two file split is less of a problem than the original, since the IP can be set to the AP name instead of to the serial/mac)
>
> I dont think including the entire address is a problem. In the case of a failure, individuals wont have to modify the IP addresses during the show only update serials.
>
> We could infer it from the switch-configurations but I think this becomes too obscure. Since now youre creating a composite unique identifier of the building + last octet.

unfortunantly switch names and AP names do not follow a similar naming 
convention, so it's harder than it should be to correlate them.

>> the other two are only going to be useful when we try to do a graphical monitoring display (and need to be combined with the coordinates within the room)
>
> This should be a separate issue since the graphical display hasnt been built.

then should we remove the map and coordinates from the file?

We have the laser tape measures this year, and intend to collect the coordinates 
within the room needed to populate the room-relative coordinates.

>> speaking of monitoring, with Zabbix, we had the APs auto-register themselves, is the new monitoring system getting pre-loaded with the devices to monitor or scanning the network to discover them?
>
> This is already taken care of by https://github.com/socallinuxexpo/scale-network/pull/649 which consumes the aps from pkgs.scaleInventory (inventory.py), builds an appropriate config and scrapes the AP prometheus endpoint.
>
>> Also, should we be creating aplist.csv? or should the things that currently use aplist.csv be modified to use the separate files?
>
> No, its no longer need. Im proposing with this PR to combine `apuse.csv` and `aps.csv` within `inventory.py`. As proposed in https://github.com/socallinuxexpo/scale-network/pull/585#issuecomment-1801099169 All of the necessary changes are within this PR. There isnt anything else that makes use of the `aplist.csv` that Im aware of.

ok

sarcasticadmin:
> how do you know what room name is in which building? that's the roommap.csv file
that was being used

We dont have that, unless we try to derive that out cross multiple files in the switch-configuration. But again lets create this in a separate issue.

> the data doesn't exist, the file contains the placeholders for them.

> unfortunantly switch names and AP names do not follow a similar naming convention, so it's harder than it should be to correlate them.

I was referring to correlating the subnet to the room so we dont have to include the subnet in the roommapping.

>  the IP can be set to the AP name instead of to the serial/mac

To clarify on your original comment. Its already being set to the AP name, the serial is then a CNAME.

Regardless I still stand by my original comment:  I dont think including the entire address is a problem. In the case of a failure, individuals wont have to modify the IP addresses during the show only update serials.

> then should we remove the map and coordinates from the file?
> We have the laser tape measures this year, and intend to collect the coordinates
> within the room needed to populate the room-relative coordinates.

Im referring to a map thats displayable along side the current monitoring solution, previous map was for zabbix.

Great, but Im still in favor of separating that work into a separate issue.

davidelang (approved):
LGTM

davidelang:
Just had one additional though, what happens if there are entries in aps that do 
not exist in apuse? do the extra serial numbers get ignored? should they get 
created as spare# or similar?

I think my preference would be that they should be ignored by inventory.py

David Lang

sarcasticadmin:
> Just had one additional though, what happens if there are entries in aps that do
not exist in apuse? do the extra serial numbers get ignored? should they get
created as spare# or similar?
>
>I think my preference would be that they should be ignored by inventory.py

Yep, they are ignored which was one of the benefits so we dont have to keep track of spares. Spares can then be calculated by the delta of anything not in `apuse.csv`

owendelong (approved):
YAY!!

Issue #587, opened 2023-03-13
Fix routing table on massflash


sarcasticadmin:
## Description
While using the massflash livecd and first associating it to another network, the user is unable to ping the flash network even after associating a nic via the addtobr

Currently you have to delete the existing default route ~~and add back one for only the massflash subnet~~ (didnt have to add anything back the last time we ran massflash):

```
ip route del default via 192.168.252.1
```

## Acceptance Criteria
- routing behaves as expecting in massflash

nixinator:
During mass flash mode , is it possible to configure the internface on the AP as a normal interface, i.e. without a VLAN.

Or do these things need to be able to flash in the field or something?

Configuring massflash without the added vlan stuff on top might be a more optimum solution.

 I'll have to study massflash a little more and see exactly how it works....

davidelang:
masflash assumes that the routers are still configured from the last show, so 
their uplink port is a trunk with them looking for a DHCP address on one of two 
VLANs (the management VLAN for each building)

when it gets an IP address, the server then runs commands via ssh/scp to upload 
the new image to the AP and flash it. This requires that the AP have a known ssh 
key/password to allow this access.

Any requirement to reconfigure the interfaces defeats the purpose of the mass 
flash and would take us back to something similar to the prior version, which 
required booting each device with the reset held down to go into debricking mode 
(where they all would come up on the same IP, so could only be done one at a 
time)

David Lang

sarcasticadmin:
> During mass flash mode , is it possible to configure the internface on the AP as a normal interface, i.e. without a VLAN.
>
> Or do these things need to be able to flash in the field or something?

@nixinator its a good question and I initially thought about the backdoor interface we put on the device as a last resort: https://github.com/socallinuxexpo/scale-network/blob/master/facts/secrets/ar71xx-openwrt-show.yaml#L97-L101 However this too has a vlan :disappointed: additonally there all set to a static address so thats not really going to meet our needs.

As @davidelang points the other option would be to reconfigure them individually which is pretty time consuming.

I think massflah pretty close to what we need as-is. It just seems like a small order of operations that leaves the routing table in a broken state.

owendelong:
We should be able to add the route changes necessary to the ifup hooks for the interface in the live image.

owendelong:
One option to deal with both circumstances might be to set up a native VLAN on the mass flash switch in addition to the tagged VLANs and have mass flash interact with both networks.

[...]

PR #680, opened 2024-03-02
[READY] - massflash routing table fix

sarcasticadmin:
## Description of PR

Fixes: #587

Routing table shouldnt set the default route to the massflash vlan. Prior to this we would have to run the following after the massflash laptop joined a wifi network to get out to the internet:

```
ip route del default via 192.168.252.1
```

## Previous Behavior
- Explicitly setting `massflash` vlan as default route

## New Behavior
- Omit any default route for the `massflash` vlan
- update docs for massflash since there were formatting errors
- .gitignores for some of the nixosTest vm images and history

## Tests
- `nix build .#nixosConfigurations.massflash.config.system.build.isoImage` and added massflash to wifi network and it was able to reach the internet without having to remove the default route

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ switch-configuration/config/switch-maps
 # Ignore nix-isms
 result
 **/*.qcow2
+**/*.nixos-test-history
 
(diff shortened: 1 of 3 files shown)


owendelong (approved):
Looks good to me. Thanks, Rob!!

PR #684, opened 2024-03-06
[READY] - massflash includes common utils

sarcasticadmin:
## Description of PR

Relates to: #587

Figured it would be a good idea to include the common utilities in our massflash livecd.

## Previous Behavior
- Basic utils were only included in `massflash` livecd

## New Behavior
- `massflash` has all our utilies from `base.nix`

## Tests

- Built livecd and tested on lenovo laptop


owendelong (approved):
LGTM

PR #690, opened 2024-03-12
Replaced confIDF with Spare

owendelong:
Routine-ish switch swap


sarcasticadmin (approved):
The snake gods thank you

PR #697, opened 2024-03-13
[WAIT] - pilist: update

MatthewCroughan:
Updates the Pi List to be accurate


owendelong (approved):
I presume that the consumers have been updated to parse the new file format?

sarcasticadmin:
I have to do some digging here but I don't think the consumers have been updated yet (hints why we left it open)

owendelong:
Sorry… It wasn’t marked WAIT or anything like that, so I thought it was ready to go.

Owen

sarcasticadmin:
Good point, Ive update the PR to include the `WAIT`

kylerisse:
thanks @MatthewCroughan

PR #702, opened 2024-03-14
[READY] update wifi 2.4 and 5 tests to check for specific approved channels

### LINKED ISSUE #701: Update wifi channel validation to enumeration
## Description
checking for a valid wifi 2.4 channel currently allows any value between 1 and 11. This should only allow 1, 6, or 11

## Acceptance Criteria
isvalidwifi24chan function only passes for channels 1, 6, or 11

kylerisse:
Fixes #701 

## Description of PR
Update the wifi data validation functions to be more restrictive.

## Previous Behavior
Data validation for 2.4 allowed any value between 1 and 11.
Data validation for 5 allowed a lot more values than it should have.

## New Behavior
Data validation for 2.4 only allows 1, 6, and 11
Data validation for 5 now allows an enumerated list.

## Tests
How was this PR tested?
- locally using pytest, validated incorrect values and also strings (due to the int casting necessary)

### DIFF
--- a/facts/datasource.py
+++ b/facts/datasource.py
@@ -68,7 +68,7 @@ def isvalidmac(macaddr):
 
 def isvalidwifi24chan(chan):
     '''test for valid 2.4Ghz WiFi channel'''
-    return int(chan) >= 1 and int(chan) <= 11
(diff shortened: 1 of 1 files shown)


davidelang (approved):
LGTM

PR #704, opened 2024-03-14
[READY] - inventory fix for aps and apsuse

sarcasticadmin:
## Description of PR

related to: https://github.com/socallinuxexpo/scale-network/pull/698

Currently failing off master when aps are all defined in `apuse.csv`:
```
$ nix build .#scaleInventory
path '/home/rherna/workspace/scale-network/nix' does not contain a 'flake.nix', searching up
error: builder for '/nix/store/0v2dvmnmngz4lislshijlgjbybs2sxqm-scaleInventory.drv' failed with exit code 1;
       last 8 log lines:
       > Traceback (most recent call last):
       >   File "/nix/store/lfs3m43yvcn3hrphicc132py43zqg9x5-scaleInventory/.repo/facts/inventory.py", line 718, in <module>
       >     main()
       >   File "/nix/store/lfs3m43yvcn3hrphicc132py43zqg9x5-scaleInventory/.repo/facts/inventory.py", line 697, in main
       >     aps = populateaps(apsfile, apusefile)
       >   File "/nix/store/lfs3m43yvcn3hrphicc132py43zqg9x5-scaleInventory/.repo/facts/inventory.py", line 299, in populateaps
       >     ipaddr = elems[2]
       > IndexError: list index out of range
       For full logs, run 'nix log /nix/store/0v2dvmnmngz4lislshijlgjbybs2sxqm-scaleInventory.drv'.
```

## Previous Behavior
- broken on `master`

## New Behavior
- fixed for scaleInventory
- include ci build as sanity check for scaleInventory

## Tests
- ci

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,3 +19,4 @@ jobs:
       - run: nix build -L .#checks.x86_64-linux.pytest-facts
       - run: nix build -L .#checks.x86_64-linux.perl-switches
       - run: nix build -L .#checks.x86_64-linux.openwrt-golden
+      - run: nix build -L .#scaleInventory
--- a/facts/inventory.py
+++ b/facts/inventory.py
@@ -296,7 +296,11 @@ def populateaps(apsfile, apusefile):
     # key: n8t-0054
     # values: c4:04:15:ad:a3:93,unknown-2,10.128.3.249,6,36,0,0,50,50
     for key, elems in result.items():
-        ipaddr = elems[2]
+        try:
+            ipaddr = elems[2]
+        except IndexError:
+            ipaddr = None
+
         # Lets bail if ip address is invalid
         if not isvalidip(ipaddr):
             continue


kylerisse (approved):
Ugh, I don’t see how I missed this and not quite clear which line isn’t valid. Will investigate tomorrow. Thanks @sarcasticadmin

sarcasticadmin:
Ya no worries, i should have caught this when i split the files originally

kylerisse:
> Ya no worries, i should have caught this when i split the files originally

I realized what the issue was while sleeping. I was only spot checking each individual file for validity and not making sure the line count across the two matched. Since that is an OK state, then yup this all makes sense now. Nice fix. I like running inventory pkg as part of CI as a sanity check, this is similar to the step in CI that piped the output of inventory through jq, back when it only spit out json.

I now have a concern about duplicate IPs (or names or mac addresses) existing in a data file, which would still produce a valid inventory pkg run but consumer services would then fail downstream. For example trying to give Kea a config that has multiple DHCP reservations for the same IP and/or mac. Will think about it some more and open an issue if it still makes sense when I get to the NOC.

PR #705, opened 2024-03-14
[READY] start using new aps

davidelang:
see diff


kylerisse (approved):
looks good, thanks @davidelang

PR #735, opened 2024-03-15
Took unused CTF switches out of switchtypes

owendelong:
## Description of PR
Blame Canada^wLuis

## Previous Behavior
CTF switches hung up the config push process

## New Behavior
CTF is out of the way.

## Tests
Well, uh, ya see... We didn't so much test it as we just figured it'd probably work.


kylerisse (approved):
🚢

PR #736, opened 2024-03-16
[READY] Fixed filter and other errors

owendelong:
## Description of PR
Packet Packet Packet --- Packet --- Packet Packet

## Previous Behavior
Switch didn't like config file error

## New Behavior
Switch will digest the config file without reflux
Also includes improvements to config loader and push to allow skipping group Z (default for push) or including it (default for build)

## Tests
Gave the switch a whole bunch of malox and threw the resulting configs at it.


kylerisse (approved):
Thanks @owendelong !

PR #740, opened 2024-03-16
[READY] remove old example configs

kylerisse:
## Description of PR
These are 6 year old configs and `inventory.py` output generated the first year of using Ansible. The goal was to build confidence among the team that configuration management was resulting in the correct output. We don't need them anymore.

## Previous Behavior
Some stale files existed

## New Behavior
The stale files don't exist anymore

## Tests
No need, they're not referenced from anywhere.


sarcasticadmin (approved):
cleanup the :wastebasket:

PR #749, opened 2024-03-17
[READY] - disable dhcpv6 for now

sarcasticadmin:
## Description of PR

Since we arent using tftp over v6 and theres a race condition with both the kea v4 and v6 daemons

## Previous Behavior

## New Behavior

## Tests
How was this PR tested?


kylerisse (approved):
Thanks @sarcasticadmin

Issue #537, opened 2023-02-18
RPi PXE Boot


MrHamel:
## Description
Add RPi PXE Booting to Automation

## Acceptance Criteria
What is the expected outcome that resolves this issue?

Be able to build a VM with the necessary config files and packages to PXE boot a number of RPi's.

GitHub is picky about file extensions used when uploading files to issues.

[exports.txt](https://github.com/socallinuxexpo/scale-network/files/10775065/exports.txt) -> /etc/exports
[kea-dhcp4.conf.txt](https://github.com/socallinuxexpo/scale-network/files/10775064/kea-dhcp4.conf.txt) -> /etc/kea/kea-dhcp4.conf

It took me awhile to get to this due to physical therapy, lost my job, and prior arrangements. I have a complete image of "server5" so I can grab anything that is needed or share the partition image.

While making the image, I've learned the hard way that Clonezilla's use of `partclone` does not produce a loop mountable partition images by default even if it's RAW type. Luckily `partclone.restore` was able to remake the file into something that can be mounted with ease.

I'm currently in the process of `xz` compressing the sda3 raw partition file as I write this, and will upload it to my server at the colo for anyone who wishes to download it.

davidelang:
following up on this, are the individual server images posted anywhere? we need 
them right away if we are going to use them this year.

David Lang

MrHamel:
Server images meaning the RPi? They are files. I was lead to believe that another NFS rootfs was in the works for the RPi's?

If you're referring to sda3 partition of "server5", I dropped the ball on getting the image uploaded, it'll be available at https://ryanhamel.com/scale/sda3-ext4.raw.xz in 90 minutes (when the SCP will complete). File is ~22GB in size, going over a 40Mbit cable upload.

I can bring the server over to the convention center tomorrow afternoon, would be best to run off that for this year, and pick back up on virtualizing it after the conference. I have been making job hunting a much higher priority.

davidelang:
the new pi image did not get created, so we want to use the one from that 
server.

we could recreate that pi image withut too much hassle, but the initrd to boot 
from is more work.

David Lang

MrHamel:
@davidelang Fixed and tested, was a permissions issue.

MrHamel:
The other partitions to go along with it:

https://ryanhamel.com/scale/sda1.dd-ptcl-img.uncomp.xz
https://ryanhamel.com/scale/sda2-vfat.raw.xz

[...]

PR #753, opened 2024-03-22
[WAIT] - [PENDING ABANDON] rpi tftpboot via dhcpv6 config

sarcasticadmin:
## Description of PR

Relates to: #537 

Captures the efforts of Tuesday and Wednesday during Scale 21x to get tftpboot working on ipv6. We still have an outstanding issue with the rpi firmware (https://github.com/raspberrypi/firmware/issues/1808) and booting from rpi4 still isnt working. At this time we cannot incorporate it into the kea dhcpv6 server since then any rpis with dhcpv6 boot enabled will fail to boot or boot into a blank screen (if tftpboot server exists). We dont want to lose the work so well leave it here for now.

## Previous Behavior
- No tftpboot configuration in kea dhcpv6

## New Behavior
- Initial tftpboot configuration for kea dhcpv6 (still WIP)

## Tests
- Debugging directly on the Rpi4
- TBD: `nix build .#checks.x86_64-linux.core.driverInteractive` works as expected for all vms
- TBD: `nix build .#checks.x86_64-linux.core` works

### DIFF
--- a/facts/inventory.py
+++ b/facts/inventory.py
@@ -504,6 +504,70 @@ def generatekeaconfig(servers, aps, vlans, outputdir):
       }
     keav6_config = {
         "Dhcp6": {
+            "option-def": [
+                  {
+                    "name": "PXEDiscoveryControl",
+                    "code": 6,
+                    "space": "vendor-40712",
+                    "type": "uint8",
+                    "array": False
+                  },
+                  {
+                    "name": "PXEMenuPrompt",
+                    "code": 10,
+                    "space": "vendor-40712",
+                    "type": "record",
+                    "array": False,
+                    "record-types": "uint8,string"
+                  },
+                  {
+                    "name": "PXEBootMenu",
+                    "code": 9,
+                    "space": "vendor-40712",
+                    "type": "record",
+                    "array": False,
+                    "record-types": "uint16,uint8,string"
+                  }
+            ],
+            "client-classes": [
+                  {
+                    "name": "rpi-pxe",
+                    "test": "option[client-arch-type].hex == 0x0029",
+                    "option-data":
+                    [
+                      {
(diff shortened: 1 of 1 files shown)


owendelong:
So some things I've learned since this was pushed up may render it generally obsolete.

For example, vendor-40712 is the vendor code for the DHCPCD project. Vendor class identifiers don't have any central registry and everyone sort of "makes them up" on a vendor-by-vendor basis (sigh).

The boot file URL if it's going to use an IPv6 address and not a name (which appears to be necessary) needs to be of the form "tftp://[ipv6_addr]/filename" or "tftp://[ipv6-addr]:port/filename", so what's in this PR won't work no matter what.

The IPv6 support in the raspberry pi boot loader is pretty thoroughly borked. The only environment it appears to work in is one where DNS is not required and the TFTP and DHCP servers are running on the same host and that host must be on-link with the Raspberry Pi.

Most of the PXE options don't appear to be understood by the R.Pi v6 boot loader, though this may require further experimentation.

For now, keep this open as a data point, but likely it should never get merged as it certainly can't work.

The long term solution might be to convert the server interface to a trunk port and plumb the necessary VLANs to the server on virtual interfaces so that TFTP and DHCP can be served on-link to the raspberry pis in question.

sarcasticadmin:
> For now, keep this open as a data point, but likely it should never get merged as it certainly can't work.

@owendelong would you be ok with me closing this and moving the branch to: `keep/rh/1710876236dhcpv6` That way we can clean up some of the cruft from the list of open PRs but its clean the branch should be kept around as a reference.

sarcasticadmin:
Moved to https://github.com/socallinuxexpo/scale-network/tree/keep/rh/1710876236dhcpv6 so its obvious to keep it around for future reference

PR #768, opened 2024-09-22
[READY] - Adding Josh to users

jshcmpbll:
## Description of PR

Adding myself to users for nix

## Previous Behavior

Josh was not a user and key was not tracked

## New Behavior

Josh user `jsh` is added and key is included for access

## Tests

na


sarcasticadmin (approved):
@jshcmpbll Thanks Josh, looks good 👍

PR #778, opened 2024-09-22
Update switch_config_loader

jeffreytjung:
Comment had load spelled wrong

## Description of PR

The comment in the top of the file was referencing a bulk_local_load_switches

## Previous Behavior

The header to the referenced part was that it was spelled with the 'o' and the 'a' were switched around

## New Behavior

The header is read with the word spelled correctly


sarcasticadmin (approved):
Thanks @jeffreytjung appreciate this fix!

Issue #781, opened 2024-09-23
substituter errors in CI (GH rate limit)


kylerisse:
## Description
Since  #770 , CI showing errors when using the magic cache substituter.

Example from https://github.com/socallinuxexpo/scale-network/actions/runs/10977769534/job/30480116742?pr=770 which is not limited to the new NixOS configuration validations, also impacts the existing checks.

```
copying path '/nix/store/qkpd5prdrc6dnqy0wpcqznr83hzn07pi-freetype-2.13.1' from 'http://127.0.0.1:37515/'...
copying path '/nix/store/bhmynyjwzc2r6iqf7fhc3yzjcv3paiwa-curl-8.3.0' from 'http://127.0.0.1:37515/'...
warning: error: unable to download 'http://127.0.0.1:37515/nar/1hdp17l0h950ym38nsjxbp4m1kvpa7jsdv0l1521q3v7mampmd21.nar.xz': HTTP error 429

       response body:

       GitHub API error: API error (429 Too Many Requests): StructuredApiError { message: "Request was blocked due to exceeding usage of resource 'Count' in namespace ''." }; retrying in 310 ms
copying path '/nix/store/4g9zd3czh9k1vy53f4mhadjqsm5ma82a-boost-1.81.0' from 'http://127.0.0.1:37515/'...
disabling binary cache 'http://127.0.0.1:37515/' for 60 seconds
error: unable to download 'http://127.0.0.1:37515/nar/109hhxp4qqdw32rp6v5ljjk7l3a9k9i9796gdmmi2swhpa23r23y.nar.xz': HTTP error 418

       response body:

       GitHub API error: GitHub Actions Cache throttled Magic Nix Cache. Not trying to use it again on this run.
copying path '/nix/store/1hpgya88k6424flhwrwfli09rq3yxb8q-nmap-7.94' from 'http://127.0.0.1:37515/'...
error: substituter 'http://127.0.0.1:37515/' is disabled
copying path '/nix/store/fmh3s032bcsbfcdp82zsjlmkj1kp72j6-sqlite-3.43.1' from 'http://127.0.0.1:37515/'...
copying path '/nix/store/4g9zd3czh9k1vy53f4mhadjqsm5ma82a-boost-1.81.0' from 'https://cache.nixos.org/'...
copying path '/nix/store/74y3751gsixaz9797ib0hp7c658sp1y5-binutils-2.40' from 'http://127.0.0.1:37515/'...
[...]

PR #783, opened 2024-09-24
[READY] - gha: limit nixos builds to master and manual

sarcasticadmin:
## Description

Fixes: #781

Running these on every PR or every push to a PR quickly gets us rate limited by github. Its unclear if theres a way around this cache rate limit:
https://github.com/DeterminateSystems/magic-nix-cache-action?tab=readme-ov-file#usage-notes

Moving this to only running on `master` and manually should do the trick for now. If it persists we might have to move it to a scheduled job like openwrt builds.

## Previous Behavior

- Bumping up against gha cache limits for many PRs due to nixosConfiguration builds

## New Behavior

- ci-nixos only running on `master` and manually

## Tests

- gha for existing PR `nix builds` jobs
- will confirm for ci-nixos jobs after they are merged to `master`


djacu (approved):
Looks like a good compromise for now

PR #784, opened 2024-10-01
[REVIEW] - Adds test for switch config types

### LINKED ISSUE #676: init datasource test for switch-config types
## Description

Currently we dont have any tests that cover the configuration for types files: https://github.com/socallinuxexpo/scale-network/tree/8cb8aab0615785177693cfe06b8f0fab51756d30/switch-configuration/config/types we should be verify the number of columns and things like `"Uplink", "Downlink", "AP", etc.` that are specific to columns. We don't have to be ultra specific per column but covering the basics for the types.

Examples of similar tests would be something like `test_switchtypes_tsv`: https://github.com/socallinuxexpo/scale-network/blob/8cb8aab0615785177693cfe06b8f0fab51756d30/facts/test_datasources.py#L80-L99

## Acceptance Criteria
- verifying the number of fields
- verify the trunktypes (e.g. "Uplink", "Downlink", "AP", etc.)

cjmakes:
## Description of PR

- Formats switch-config/type so that each config is a single line
- Adds a datasources test for switchconfig types

## Previous Behavior

There was no automation to enforce consistency across switch config types.

## Tests

This pr adds tests. I would appreciate feed back on how to make sure I didn't break any automation that parses the switch configs and relies on using \ to split a config across lines.

closes #676

### DIFF
--- a/facts/test_datasources.py
+++ b/facts/test_datasources.py
@@ -107,6 +107,7 @@ def test_switchtypes_tsv():
     result, err = ds.test_tsvfile(meta)
     assert result, err
 
+
 def test_switchconfigs_tsv():
     """test switchconfigs"""
 
@@ -127,6 +128,7 @@ def test_switchconfigs_tsv():
         }
         result, err = ds.test_tsvfile(meta)
         assert result, err
+
 
 def test_vlansd_tsv():
     """test vlans.d/"""
--- a/facts/test_datasources.py
+++ b/facts/test_datasources.py
@@ -121,7 +121,7 @@ def test_switchconfigs_tsv():
                 ds.isuntested,
                 ds.isuntested,
                 ds.isuntested,
-                ds.isuntested,
+                ds.isvalidlink,
                 ds.isuntested,
             ],
         }
--- a/facts/test_datasources.py
+++ b/facts/test_datasources.py
@@ -107,6 +107,26 @@ def test_switchtypes_tsv():
     result, err = ds.test_tsvfile(meta)
     assert result, err
 
+def test_switchconfigs_tsv():
+    """test switchconfigs"""
+
+    directory = "../switch-configuration/config/types/"
+    for filename in os.listdir(directory):
+        meta = {
+            "file": directory + filename,
+            "header": False,
+            "count": "1+",
+            "cols": [
+                ds.isvalidport,
+                ds.isuntested,
+                ds.isuntested,
+                ds.isuntested,
+                ds.isuntested,
+                ds.isuntested,
+            ],
+        }
+        result, err = ds.test_tsvfile(meta)
+        assert result, err
 
 def test_vlansd_tsv():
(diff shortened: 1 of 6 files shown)


sarcasticadmin (approved):
thanks for push this up @cjmakes ! Ill review it more in depth when I have some time later this week, but lets approve it so we can see how CI runs

also tagging @owendelong since this changes the formatting of the switchtypes files slightly

facts/test_datasources.py
```diff
@@ -108,6 +108,28 @@ def test_switchtypes_tsv():
...
 
+def test_switchconfigs_tsv():
+    """test switchconfigs"""
+
+    directory = "../switch-configuration/config/types/"
+    for filename in os.listdir(directory):
+        meta = {
+            "file": directory + filename,
+            "header": False,
+            "count": "1+",
```
sarcasticadmin, line 119: I think the count here should be `4+` since it looks like we have a mix of `4-6` columns for the config types

facts/test_datasources.py
```diff
@@ -108,6 +108,28 @@ def test_switchtypes_tsv():
...
+            "cols": [
+                ds.isvalidport,
```
sarcasticadmin, line 121: having `isvalidport` and `isvalidlink` is definitely a great start to testing these switch config types, nice work!

switch-configuration/config/types/Booth
```diff
@@ -1,8 +1,6 @@
 // Expo Booth Area switch Template
-TRUNK	ge-0/0/0	exSCALE-SLOW,exSCALE-FAST,exSpeaker,exInfra,exAVLAN, \
-				exSigns,exVmVendor,vendor_backbone			-	Uplink
-TRUNK	ge-0/0/1	exSCALE-SLOW,exSCALE-FAST,exSpeaker,exInfra,exAVLAN, \
-				exSigns,exVmVendor,vendor_backbone			-	Downlink
+TRUNK	ge-0/0/0	exSCALE-SLOW,exSCALE-FAST,exSpeaker,exInfra,exAVLAN, exSigns,exVmVendor,vendor_backbone			-	Uplink
```
sarcasticadmin, line 2: Id prefer it we update `test_datafile` to support the `\` for multiline entries. Would you mind seeing how feasible that approach would be?

sarcasticadmin (commented):
@cjmakes this looks really good! Left my comments inline so let me know what you think

sarcasticadmin:
Thanks for fixing the linting errors that came up in CI as well 👍

kylerisse (changes requested):
still awaiting multi-line support. i say we just close this for now.

sarcasticadmin:
@cjmakes were going to close this for now since it seems like there are a few limitations with how parse these switch-configuration files after I touched based with @kylerisse 

We can revisit after we have a better plan for approaching the processing of the multiline switch configs and types.

PR #498, opened 2022-07-30
[READY]  - fixing apinger and massflash updates

sarcasticadmin:
## Description of PR

apinger updates, dhcpclient reload/restart services, and massflash being used to update the APs

## Previous Behavior
- `apinger` template wasnt working when no network was present
- `dhcpclient` reboots when hostname changes
- AP will yoyo if it has amnesia of its hostname

## New Behavior
- Fixing `apinger` on boot when no network present at first
- `dhcpclient` reloads services and system when hostname changes
- AP has stable uptime stable
- `massflash` doesnt have a hardcode sha

## Tests
- [x] - Loaded it on `flash1` correctly
- [x] - Test cases: flash -> bring up without network results in down the radios
- [x] - Test cases: unplug network and replug with radios going down then up as expected 
- [x] - serverspec tests pass
- [x] - load on `noc1` and AP has been running successfully since

PR #785, opened 2024-10-08
[REVIEW] - udhcp.user uci commit only new hostnames

### LINKED ISSUE #442: udhcp.user unnecessary commits
## Description
udhcp.user commits hostname whether or not it has changed.

## Acceptance Criteria
Test if old hostname and DHCP supplied hostnames are identical and don't commit if no change.

ruebenramirez:
## Description of PR

Instead of always committing a provided hostname, only commit new hostnames, where the new hostname doesn't match the current hostname.

Resolves #442 

## Previous Behavior

Any time a hostname value was available, `udhcp.user` would `uci commit` the new hostname.

## New Behavior

`udhcp.user` will only `uci commit` the hostname if it is different than the current hostname.

## Tests

(need some help here)
- [ ] test against AP hardware (monitor for `uci commits`)
- [ ] provide the same hostname change to the device
- [ ] provide a new hostname to the device

### DIFF
--- a/openwrt/files/etc/udhcpc.user
+++ b/openwrt/files/etc/udhcpc.user
@@ -21,17 +21,18 @@ case "$1" in
       /root/bin/apinger-pop.sh "$router"
     fi
 
-    if [ ! -z "$hostname" ]; then
+    if [ ! -z "$hostname" ] && [ `echo "$HOSTNAME" | tr '[A-Z]' '[a-z]'` != `echo "$hostname" | tr '[A-Z]' '[a-z]'` ]; then
+      # set new hostname
       uci set 'system.@system[0].hostname'="$hostname"
       uci commit
-      if [ `echo "$HOSTNAME" | tr '[A-Z]' '[a-z]'` != `echo "$hostname" | tr '[A-Z]' '[a-z]'` ]; then
-        # reload/restart whatever needs the hostname updated
-        /etc/init.d/system reload
-        service zabbix_agentd restart
-        service rsyslog restart
-        service lldpd restart
-      fi
+
+      # reload/restart whatever needs the hostname updated
+      /etc/init.d/system reload
+      service zabbix_agentd restart
+      service rsyslog restart
+      service lldpd restart
     fi
+
     if [ ! -z "$opt226" ]; then
       /root/bin/config-version.sh -c $(printf %d "0x$opt226")
     fi
--- a/tests/unit/openwrt/golden/ar71xx/etc/udhcpc.user
+++ b/tests/unit/openwrt/golden/ar71xx/etc/udhcpc.user
@@ -21,17 +21,18 @@ case "$1" in
       /root/bin/apinger-pop.sh "$router"
     fi
 
-    if [ ! -z "$hostname" ]; then
+    if [ ! -z "$hostname" ] && [ `echo "$HOSTNAME" | tr '[A-Z]' '[a-z]'` != `echo "$hostname" | tr '[A-Z]' '[a-z]'` ]; then
+      # set new hostname
       uci set 'system.@system[0].hostname'="$hostname"
       uci commit
-      if [ `echo "$HOSTNAME" | tr '[A-Z]' '[a-z]'` != `echo "$hostname" | tr '[A-Z]' '[a-z]'` ]; then
-        # reload/restart whatever needs the hostname updated
-        /etc/init.d/system reload
-        service zabbix_agentd restart
-        service rsyslog restart
-        service lldpd restart
-      fi
+
+      # reload/restart whatever needs the hostname updated
+      /etc/init.d/system reload
+      service zabbix_agentd restart
+      service rsyslog restart
+      service lldpd restart
     fi
+
     if [ ! -z "$opt226" ]; then
       /root/bin/config-version.sh -c $(printf %d "0x$opt226")
     fi
--- a/tests/unit/openwrt/golden/ipq806x/etc/udhcpc.user
+++ b/tests/unit/openwrt/golden/ipq806x/etc/udhcpc.user
@@ -21,17 +21,18 @@ case "$1" in
       /root/bin/apinger-pop.sh "$router"
     fi
 
-    if [ ! -z "$hostname" ]; then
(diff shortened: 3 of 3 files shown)


openwrt/files/etc/udhcpc.user
```diff
@@ -22,9 +22,11 @@ case "$1" in
     fi
 
     if [ ! -z "$hostname" ]; then
-      uci set 'system.@system[0].hostname'="$hostname"
-      uci commit
       if [ `echo "$HOSTNAME" | tr '[A-Z]' '[a-z]'` != `echo "$hostname" | tr '[A-Z]' '[a-z]'` ]; then
```
ruebenramirez, line 25: I suppose we could combine the multiple conditional statements here, but this seems more readable than that would be.

kylerisse: haha, more unreadable than it is already? 

```
if [[ ! -z "$hostname" ] && [ `echo "$HOSTNAME" | tr '[A-Z]' '[a-z]'` != `echo "$hostname" | tr '[A-Z]' '[a-z]'` ]];
```

I think this change makes sense. Originally this script would reboot if the hostname changed, but that was subsequently replaced with a service reload as you can see. I still think you want the `[ ! -z "$hostname" ]` check for cases when DHCP server doesn't supply this. I can't actually think of how this would happen, but stranger things have occurred.

@sarcasticadmin it's been years since I thought about this file. Is the understanding correct? We should only `uci set` the new hostname if it changes rather than every renew and 100% of the time that is the case we should restart the services. I'm not clear on why it's being set every time, but you might have been doing this for a reason.

sarcasticadmin: I like the compound `if` @kylerisse suggested since we should handle the case where `hostname` is not set from the dhcp server side (kea) even though its unlikely.

As for why this condition exists at all: we set the hostname to the value from the `name` field in https://github.com/socallinuxexpo/scale-network/blob/ef74408a99827005e7e36f5b2b5ab78784611f1e/facts/aps/apuse.csv#L1 It was originally identified around ~Scale 19x or so that kea would push a lower case hostname: `openwrt` while by default the hostname in our builds was `OpenWRT`. Upon commit and reboot that name didnt always take (persistent memory issues on the AP) even though it was technically equivalent minus the capitalization: https://github.com/socallinuxexpo/scale-network/pull/441 covers the justification, fix, and some additional finds. 

Later we opted to restart the services that needed the hostname update instead of the whole AP so the APs dont get in a reboot loop: https://github.com/socallinuxexpo/scale-network/pull/498/commits/2d785b30b92d25cf839456a04d1a1668c1c25332

Some of this is a little fuzz but hopefully that explains why we do this.

sarcasticadmin (commented):
@ruebenramirez I think this all makes sense, the one suggestion from @kylerisse for making a compound `if` seems like its worthwhile.

Regarding testing this, I can circle back with you once you have time about getting it running through the real hardware CI via gitlab.

openwrt/files/etc/udhcpc.user
ruebenramirez: Wow, thank you both so much for the context and feedback!

ruebenramirez:
> @ruebenramirez I think this all makes sense, the one suggestion from @kylerisse for making a compound `if` seems like its worthwhile.
> 
> Regarding testing this, I can circle back with you once you have time about getting it running through the real hardware CI via gitlab.

Compound conditional in in place now.  I'm looking forward to learning more about the hardware testing process.

openwrt/files/etc/udhcpc.user
kylerisse: > I like the compound if @kylerisse suggested

Ha! @ruebenramirez suggested this, I just wrote it out to see how unreadable it would be. Seems fine, typical Bash.

kylerisse:
> > @ruebenramirez I think this all makes sense, the one suggestion from @kylerisse for making a compound `if` seems like its worthwhile.
> > Regarding testing this, I can circle back with you once you have time about getting it running through the real hardware CI via gitlab.
> 
> Compound conditional is in place now. I'm looking forward to learning more about the hardware testing process.

https://github.com/socallinuxexpo/scale-network/actions/runs/11266376404/job/31329957560?pr=785 ci choking on script @ruebenramirez . I think you just need to wrap the entire compound if in another set of `[ ]`

ruebenramirez:
Derp!  Thanks @kylerisse!  Those should all be accounted for now.  If we want the extra `[ ]`'s I can throw them in, I just didn't see them in other compound conditionals.  Maybe I need to look harder though? ;)

kylerisse:
> Derp! Thanks @kylerisse! Those should all be accounted for now. If we want the extra `[ ]`'s I can throw them in, I just didn't see them in other compound conditionals. Maybe I need to look harder though? ;)

Not your derp. It's a weird thing we do. In this case it was stopping at `nix build -L .#checks.x86_64-linux.openwrt-golden`. I believe we can add you to the repo so that your checks run automatically. @sarcasticadmin has to put in a request with TPTB. CI passes now, I think we just wait for you guys to do hardware testing. And thank you for fixing up the commit!

sarcasticadmin:
Thanks @ruebenramirez for all the fixups and @kylerisse for your feedback. I request rueben be add to the @socallinuxexpo/tech so that should make this process and future openwrt testing easier 🙂 

@ruebenramirez and I will sync up on testing this on real hardware before we merge

ruebenramirez:
Hey look at me!  
image

I confirmed I can trigger workflows now too :)

kylerisse (approved):

owendelong:
If this works, that's awesome, but last time we tried to do this, we ran into some weird behavior in OpenWRT that was causing it not to commit new hostnames about 50% of the time. David might remember more details than I do. That was why we ended up implementing the always-commit behavior IIRC.

davidelang:
The existing code has this line that checks if the hostname is different:

      if [ `echo "$HOSTNAME" | tr '[A-Z]' '[a-z]'` != `echo "$hostname" | tr '[A-Z]' '[a-z]'` ]; then

davidelang:
As part of #823 we may add a setting to this file to make comparisons case insensitive, that would allow elimination of the tr commands.

So this change is a minor cleanup that I don't object to, but I think we should revisit it after this scale and after 823 is added

PR #786, opened 2024-10-10
[READY] - devServer disko config

sarcasticadmin:
## Description of PR

Instead of leveraging something like https://github.com/socallinuxexpo/scale-network/blob/ef74408a99827005e7e36f5b2b5ab78784611f1e/nix/machines/bootstrap/bootstrap.sh this moves our provisioning of systems to disko.

This configuration was used to setup the devServer utilizing the 500GB disks in a mirror.

## Previous Behavior

- Previous disks were utilizing the 30 GB SSDs

## New Behavior

- Leverage 500 GB disks for devServer usage

## Tests

- `devServer` is online

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,5 +1,26 @@
 {
   "nodes": {
+    "disko": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs-unstable"
+        ]
+      },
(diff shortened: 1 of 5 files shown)


kylerisse (approved):
Awesome!

PR #792, opened 2024-11-24
Add Juniper root passwd SHA-256 Hash

MrHamel:
## Description of PR

Add Juniper root passwd SHA-256 Hash, to simplify switch config builds.

## Previous Behavior

The file had to be created before generating switch configs.

## New Behavior

This will help in automatically generating switch configs, to leverage Juniper's event handling and ZTP (zero touch provisioning mechanisms), to push configs and firmware to devices.


sarcasticadmin (approved):
@MrHamel yep this makes sense given the existing sha elsewhere

PR #796, opened 2024-11-24
adding Carlyn's key

carlynlee:
## Description of PR

Adding key for carlyn's laptop

## Previous Behavior

did not authenticate carlyn's laptop

## New Behavior

should authenticate carlyn's laptop

## Tests

was not tested


owendelong:
Technically, it's authenticating Carly, not Carly's laptop. Host authentication is mostly automatic and not restricted. Person authentication/authorization is what we care about here.

owendelong (approved):
Appears reasonable.

owendelong:
Carlyn, you should add yourself to miniconfig as well, so please submit an additional PR for that.

Issue #801, opened 2024-11-25
document and incorporate the perl ping module patch


sarcasticadmin:
## Description

per @owendelong message in signal there is a patch that should be incorporated into the ping perl module (https://metacpan.org/pod/Net::Ping). Lets document this for others and we can also incorporate this into the patch within our `nix shell`.

Patch:

```
% diff dist_[Ping.pm](http://ping.pm/) od_[Ping.pm](http://ping.pm/) 
230c230
<     croak("icmp ping requires root privilege") if !_isroot();
---
>     croak("icmp ping requires root privilege") if !_isroot() && ($^O ne "linux");
235,236c235,244
<     socket($self->{fh}, PF_INET, SOCK_RAW, $self->{proto_num}) ||
<       croak("icmp socket error - $!");
---
>     if ($^O eq "linux")
>     {
> 	    socket($self->{fh}, PF_INET, SOCK_DGRAM, $self->{proto_num}) ||
> 	      croak("icmp socket error - $!");
>     }
>     else
>     {
> 	    socket($self->{fh}, PF_INET, SOCK_RAW, $self->{proto_num}) ||
> 	      croak("icmp socket error - $!");
>     }
253,254c261,270
<     socket($self->{fh}, $AF_INET6, SOCK_RAW, $self->{proto_num}) ||
<       croak("icmp socket error - $!");
---
>     if ($^O eq "linux")
>     {
> 	    socket($self->{fh}, $AF_INET6, SOCK_DGRAM, $self->{proto_num}) ||
> 	      croak("icmp socket error - $!");
>     }
>     else
>     {
> 	    socket($self->{fh}, $AF_INET6, SOCK_RAW, $self->{proto_num}) ||
> 	      croak("icmp socket error - $!");
>     }
```

## Acceptance Criteria

- patch is more widely know and documented for the teams
- patch incorporated into default nix shell

djacu:
This should be trivial to add to the overlays once #800 is merged.

Need more info from @owendelong about which package needs patching. I'm not able to infer it from the patch above.

sarcasticadmin:
> This should be trivial to add to the overlays once #800 is merged.
> 
> Need more info from @owendelong about which package needs patching. I'm not able to infer it from the patch above.

[...]

PR #804, opened 2024-12-01
[READY] - perl packages for devShell.networkSub to run switch_config_loader

sarcasticadmin:
## Description of PR

Fixes: #801

This requires all of the perl libs to be able to run `switch_config_loader` from inside the devShell.networkSub

## Previous Behavior

- No perl lib deps existed in devShell.networkSub

## New Behavior

- perl lib deps exist in devShell.networkSub
- perl lib Net::Ping patch included in scale-network overlay

## Tests

- TBD

### DIFF
--- a/nix/packages/perl-net-interface/package.nix
+++ b/nix/packages/perl-net-interface/package.nix
@@ -2,12 +2,13 @@
   perlPackages,
   fetchurl,
   linuxHeaders,
+  which
 }:
 
 perlPackages.buildPerlPackage {
   pname = "NetInterface";
   version = "1.016";
-  buildInputs = [ linuxHeaders ];
+  buildInputs = [ linuxHeaders which ];
   doCheck = false;
   src = fetchurl {
     url = "mirror://cpan/authors/id/M/MI/MIKER/Net-Interface-1.016.tar.gz";
--- a/nix/packages/perl-net-interface/package.nix
+++ b/nix/packages/perl-net-interface/package.nix
@@ -1,29 +1,17 @@
 {
   perlPackages,
   fetchurl,
-  linuxHeaders,
-  which
 }:
-
 perlPackages.buildPerlPackage {
   pname = "NetInterface";
   version = "1.016";
-  buildInputs = [ linuxHeaders which ];
-  doCheck = false;
-
-  CPPFLAGS="-Iusr/include";
-
-  postPatch = ''
-    mkdir -p usr/include/sys/
-    cp -pr ${linuxHeaders}/include usr/
-    ln -s ${linuxHeaders}/include/linux/socket.h usr/include/sys/socket.h
-    ls -lahR usr/
-    echo "aaaaaaaa"
-    echo $CPPFLAGS
-  '';
 
   src = fetchurl {
     url = "mirror://cpan/authors/id/M/MI/MIKER/Net-Interface-1.016.tar.gz";
     hash = "sha256-e+RGk14BPQ7dPTcfkvuQo7s4q+mVYoidgXiVMSnlNQg=";
   };
+
+  patches = [ ./net-interface-remove-grep-af-inet.patch ];
+
+  doCheck = false;
 }
--- a/nix/packages/perl-net-interface/package.nix
+++ b/nix/packages/perl-net-interface/package.nix
@@ -1,0 +1,16 @@
+{
+  perlPackages,
+  fetchurl,
+  linuxHeaders,
+}:
+
+perlPackages.buildPerlPackage {
+  pname = "NetInterface";
+  version = "1.015";
+  buildInputs = [ linuxHeaders ];
+  doCheck = false;
+  src = fetchurl {
+    url = "mirror://cpan/authors/id/M/MI/MIKER/Net-Interface-1.015.tar.gz";
+    hash = "sha256-x6MFjTGh73k+eAtqTJCQEI/SxMNFeQ774AeFKgIvf4E=";
+  };
+}
--- a/nix/packages/perl-net-interface/package.nix
+++ b/nix/packages/perl-net-interface/package.nix
@@ -6,11 +6,11 @@
 
 perlPackages.buildPerlPackage {
   pname = "NetInterface";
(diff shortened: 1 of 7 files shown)


sarcasticadmin:
@djacu Im currently having an issue with building `scale-network.perlNetInterface`. Ive tried a couple of things but the build always wants to look into `/usr/include/sys`. There are additional details around what we'd need to do in an upstream open issue: https://rt.cpan.org/Public/Bug/Display.html?id=124582 Seems like we'd need to include the old `sys` path symlinked to wherever `socket.h` resides but I can't seem to get the `configure` to look anywhere but `/usr/include/sys`. See: https://github.com/socallinuxexpo/scale-network/pull/804/commits/0290733bb867a728e4f9c589258c9208e350fae9

Heres the error Im getting when going from 1.015->1.016. Let me know if you (or anyone else for that matter) have any ideas:

```
...
checking for LOCAL_SIZEOF_SOCKADDR_EON... not defined
checking for LOCAL_SIZEOF_SOCKADDR_INARP... not defined
checking for LOCAL_SIZEOF_SOCKADDR_ISO... not defined
checking for LOCAL_SIZEOF_SOCKADDR_NS... not defined
checking for LOCAL_SIZEOF_SOCKADDR_ATM... not defined
checking for LOCAL_SIZEOF_SOCKADDR_DECnet... not defined
checking for LOCAL_SIZEOF_SOCKADDR_TP... not defined
checking for LOCAL_SIZEOF_SOCKADDR_OSITP... not defined
configure: creating ./config.status
config.status: creating config.h
/nix/store/3jc4m5gwimj4mbm01ijgkm2di8hiy5l3-gnugrep-3.11/bin/grep: /usr/include/*: No such file or directory
/nix/store/3jc4m5gwimj4mbm01ijgkm2di8hiy5l3-gnugrep-3.11/bin/grep: /usr/include/*: No such file or directory
/nix/store/3jc4m5gwimj4mbm01ijgkm2di8hiy5l3-gnugrep-3.11/bin/grep: /usr/include/*: No such file or directory
Uncaught exception from user code:
        AF_INET not found in netsymbolsC.inc,
        prerequisite development library header files missing from /usr/include/sys
```

nix/packages/perl-net-interface/package.nix
```diff
@@ -0,0 +1,29 @@
...
+  doCheck = false;
+
+  CPPFLAGS="-Iusr/include";
+
+  postPatch = ''
+    mkdir -p usr/include/sys/
+    cp -pr ${linuxHeaders}/include usr/
+    ln -s ${linuxHeaders}/include/linux/socket.h usr/include/sys/socket.h
+    ls -lahR usr/
+    echo "aaaaaaaa"
```
djacu, line 21: You are seen
You are heard
Your feelings are valid.

djacu:
> @djacu Im currently having an issue with building `scale-network.perlNetInterface`. Ive tried a couple of things but the build always wants to look into `/usr/include/sys`. There are additional details around what we'd need to do in an upstream open issue: https://rt.cpan.org/Public/Bug/Display.html?id=124582 Seems like we'd need to include the old `sys` path symlinked to wherever `socket.h` resides but I can't seem to get the `configure` to look anywhere but `/usr/include/sys`. See: [0290733](https://github.com/socallinuxexpo/scale-network/commit/0290733bb867a728e4f9c589258c9208e350fae9)

I tried some things that did not pan out. @ConnorBaker is coming over to hack in a bit. He might have some good ideas.

sarcasticadmin:
@djacu also included another package: `package.perlNetPing` which incorporates the patch provide by Owen in #801 just FYI since that wasnt in the original PR

djacu:
@sarcasticadmin I have been digging around and I'm not seeing `AF_INET` defined in any header file in `linuxHeaders` like we expect. See [this](https://students.mimuw.edu.pl/SO/Linux/Kod/include/linux/socket.h.html) where `AF_INET` is explicitly defined. I created a patch that removes the check for this symbol and it did finish building... but I don't know if this would actually be usable.

sarcasticadmin:
@djacu good catch, I should have actually looked at the contents of `linuxHeaders` :facepalm:

>  I created a patch that removes the check for this symbol and it did finish building... but I don't know if this would actually be usable.

Seems worth a shot, Im assuming you patched out the following: https://metacpan.org/release/MIKER/Net-Interface-1.016/source/Makefile.PL#L179-183 Feel free to push up what you've got too.

Im going to ping @owendelong and see if I can confirm which version hes using and on what platforms. Seems like this would fail for him on various versions of modern linux.

djacu:
> Seems worth a shot, Im assuming you patched out the following: https://metacpan.org/release/MIKER/Net-Interface-1.016/source/Makefile.PL#L179-183 Feel free to push up what you've got too.

Yeah that is what I patched out. Cleaned up the derivation definition and pushed up what I got. I tried to enable the tests but they failed and I couldn't figure out in a short amount of time how to disable only the broken tests.

It builds now but still not sure if it actually works at runtime.

nix/packages/perl-net-interface/package.nix
owendelong: Do we want to copy them or perhaps it would make more since to symlink them? Just asking, not sure which is preferable.

sarcasticadmin: were opting to patch out the test: https://github.com/socallinuxexpo/scale-network/blob/3179981d34834933f1886d1f04a8d50e7698b732/nix/packages/perl-net-interface/net-interface-remove-grep-af-inet.patch

Were not sure if that'll work or not but we can debug at the work party

sarcasticadmin:
@djacu I feel pretty good about this PR after testing it at the work party this past weekend? What do you think? Going to say this is as ready as it can be until we get another change to test against real hardware at the Expo.

djacu (approved):
If you think it's good to go then ship it. We can modify and update at con.

sarcasticadmin:
> If you think it's good to go then ship it. We can modify and update at con.

Agreed :shipit:

PR #810, opened 2025-01-08
Updated for SCaLE 22x -- Now annual files can be easily updated for future years

owendelong:
## Description of PR

T-Shirt work for 22x

## Previous Behavior

Older T-Shirt designs

## New Behavior

Stylin'!!

## Tests

Mark one eyeball utilized to test aesthetic properties to non-exacting standards.


sarcasticadmin (approved):
awesome, thanks for getting this together Owen!

graphics/current-annual Front.svg
sarcasticadmin, line 1: It should be `penguins` with a lower case p: https://github.com/socallinuxexpo/scale-network/blob/9aa9bc76ea80dd81e5a8de55bd5732f04b4f9ea1/facts/secrets/ar71xx-openwrt-show.yaml#L116

Best to keep it the same as years past so theres no confusion (old shirts and old config from guests laptops) and we wont have to update the signs server either: https://github.com/socallinuxexpo/scale-signs/blob/38fda0995cb9ec4b304dba8bc4c7dcb846b0737b/server/images/WiFi-Sign.png

sarcasticadmin (changes requested):
Found one small tweak before we send them out to get printed :sweat_smile:

graphics/current-annual Front.svg
sarcasticadmin: Thanks @owendelong for the update

sarcasticadmin (approved):

Issue #382, opened 2021-07-23
Update all openwrt refs from ar71xx to ath79


sarcasticadmin:
## Description

Relates to: #119, #369

As for August 2020 `ar71xx` build targets have been removed: https://git.openwrt.org/?p=openwrt/openwrt.git;a=commitdiff;h=4e4ee4649553ab536225060a27fc320bf54e458c

Whats weird is Im not quick sure why our version bump worked in this case: https://github.com/socallinuxexpo/scale-network/pull/374 I would have expected it to fail but maybe openwrt just aliases the `TARGET`? Either way I dont think we should depend on this since the `ar71xx` `TARGET` is clearly deprecated and called out in the docs: https://openwrt.org/docs/techref/targets/ar71xx

As followup to the openwrt `TARGET` changes made via #119, we also need to update all references to `ar71xx` during build time and inferences based on this. 

## Acceptance Criteria
Off the top of my head that includes the following things that need to be updated:

- [ ] - Makefile references that use `TARGET=ar71xx`
- [ ] - Openwrt architecture specific configs that are tied to the arch
- [ ] - Test cases that are architecture specific
- [ ] - CI tests and builds still work as intended

Issue #808, opened 2024-12-31
bump openwrt upstream version to HEAD


sarcasticadmin:
## Description

Time to update the openwrt version to catch up with upstream.

## Acceptance Criteria

- openwrt build continues to work on existing support architectures

PR #814, opened 2025-01-26
[REVIEW] - openwrt/opkg version bump and ath79 arch rename

sarcasticadmin:
## Description of PR

Relates to: #808
Fixes: #382

Version bump in preparation for SCALE 22x

## Previous Behavior

- openwrt and opkg havent been bumped since ~2023
- `ar71xx` target for netgear hardware

## New Behavior

- Bumping openwrt (sept-2024) and opkg versions (dec-2024) respectively. I will probably have a follow up for openwrt version in a future PR but would like to get this merged since its a good waypoint
- opkg is still forked due to https://github.com/openwrt/packages/issues/20184
- moving to ath79 target for the netgear hardware. This simplied the version bump since `ar71xx` was already being aliases to `ath79`
- Bumping docker image due to toolchain deps for newer versions of openwrt

## Tests

- Builds for `ath79` arch
- TBD


davidelang:
I don't disagree with this, but I think the current master has moved from opkg 
to apkg

David Lang

sarcasticadmin:
> I don't disagree with this, but I think the current master has moved from opkg to apkg David Lang

yes, but its the same repo still it just produces apkgs instead of opkgs from the same source.

PR #831, opened 2025-02-03
ar71xx: improve flashing instructions

djacu:
## Description of PR

Updated the ar71xx (ath79) instructions and account for some of the new tools (`make-dhcpd`)


sarcasticadmin (approved):
These should have all the missing bits added @djacu

Going to merge so that these instructions are available for the show

PR #800, opened 2024-11-25
[READY] nix.treewide: remove flake-parts

djacu:
## Description of PR

Removes flake parts and restructures the nix sections of the repository.

## Previous Behavior

Absolute chaos

## New Behavior

Bringing structure and order to the architecture of the nix code.
The hope is that this PR is effectively a no-op and the packages, NixOS configurations, and checks are the same.
Testing in comments below found that the only remaining differences are NixOS system closures revolving around:
- `berkhan` being added as a user to all systems
- system packages being different which is very likely attributed to `berkhan` being added (e.g. a single new instance of bash showed up) OR system packages being strings and order had changed for no reason (using `jd` confirmed that the package set was identical in the few cases that were tested)
- the hostname changed because now there is a default for all hosts 

## Tests

`nix flake check`
Lots more see below

PR #833, opened 2025-02-03
meta: architecture improvements [WIP]

djacu:
## Description of PR

Continuing the insanity of #800. Re-organizing the nix infrastructure and simplifying derivations to use standard practices.

## Previous Behavior

n/a

## New Behavior

Hopefully the same as the old behavior. :crossed_fingers: 

## Tests

`nix flake check`
Needs verification from the team that things work as expected.

### DIFF
--- a/nix/checks/default.nix
+++ b/nix/checks/default.nix
@@ -1,12 +1,10 @@
 inputs:
 let
+
   # inherits
+
   inherit (inputs.nixpkgs)
     lib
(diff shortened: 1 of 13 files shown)


sarcasticadmin:
@djacu are we ok to close this? I know you mentioned adding improvements to the nix architecture but it seemed like it would be net new, is that correct?

PR #835, opened 2025-02-03
[READY] - Instrumented Loader.pm use of Ping and Interfaces calls

owendelong:
## Description of PR

Instrumented Loader.pm

## Previous Behavior

Loader.pm was less informative

## New Behavior
Loader.pm tells you more about what's happening in its early stages

## Tests

1. Wet finger.
2. Hold finger in wind
3. First side of finger to dry indicates direction from which wind is blowing

### DIFF
--- a/switch-configuration/config/scripts/Loader.pm
+++ b/switch-configuration/config/scripts/Loader.pm
@@ -166,15 +166,19 @@ sub new
         
         STDERR->autoflush(1);   # Turn on autoflush for STDERR
         get_switchtype("anonymous");
-	
+
+        my $ping = Net::Ping->new("icmp"),
+        print STDERR "Ping OK\n";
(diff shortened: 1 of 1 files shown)


switch-configuration/config/scripts/Loader.pm
```diff
@@ -126,11 +126,15 @@ sub new
         STDERR->autoflush(1);   # Turn on autoflush for STDERR
         get_switchtype("anonymous");
 	
+            my $ping = Net::Ping->new("icmp"),
```
sarcasticadmin, line 129: @owendelong can you fix the indention? Looks off

sarcasticadmin: fixed and fixedup the commit. Mixing of tabs and spaces

PR #837, opened 2025-02-17
Update configuration.nix with Loki Service

jeffreytjung:
This is the Loki service for ingesting logs for the monitoring section of the infrastrucutre.

This change adds a loki-config.yaml file.

The commit does not have Loki fully functioning.


jeffreytjung (commented):
spelled filesystem without the m

PR #840, opened 2025-02-20
Update configuration.nix with promtail

jeffreytjung:
Promtail is a part of the monitoring stack because it is the processor for logs. 

Promtail will send logs to a Loki service which is what injests the logs. Both services are needed, but promtail goes in front of Loki. It is the point where all of the rsyslog services that are running on the access points will send their logs. Basically, send logs to Promtail’s port, not directly to Loki. It is important to note that promtail only accepts newer RFC 5424 ("IETF") formatted syslog messages.

The Grafana dashboard will then be able to have Loki presets loaded into modules.

This push is just a build that puts the promtail servicee on the local NixOS host.

### DIFF
--- a/nix/nixos-configurations/monitor/configuration.nix
+++ b/nix/nixos-configurations/monitor/configuration.nix
@@ -102,7 +102,37 @@ in
         ];
       };
     };
-
+    
+    services.promtail = {
+      enable = true;
+      configuration = {
+        server = {
+          http_listen_port = 3300;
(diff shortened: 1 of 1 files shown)


owendelong (approved):
Seems reasonable, but someone else will need to resolve the merge conflicts.

kylerisse (changes requested):
this is stale and might not align with new monitoring setup, don't merge

PR #843, opened 2025-02-25
nixosModules.monitoring: enable nginx and add tests

djacu:
## Description of PR

Use `mkDefault` for enable flags for easier toggling.
Remove `grafanaDomain` option and inline `localhost`.
Enable nginx routing and fix module definition.
Add test to check routing for grafana.

## Previous Behavior

We didn't enable nginx for monitoring.

## New Behavior

We enabled nginx for monitoring.

## Tests

`nix build .#checks.x86_64-linux.monitor`

### DIFF
--- a/nix/checks/monitor.nix
+++ b/nix/checks/monitor.nix
@@ -13,13 +13,30 @@
     virtualisation.graphics = true;
   };
 
-  testScript = ''
-    start_all()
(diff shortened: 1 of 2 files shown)


sarcasticadmin (approved):
Worked with @djacu on getting nginx configured with grafana. This doesnt enable TLS yet but leveraging routing from the header

PR #845, opened 2025-02-27
bump flake inputs & add wasgeht [DO NOT MERGE]

djacu:
## Description of PR

Depends on #846 

- Bumps the nixpkgs-unstable flake input.
- The general nixpkgs input is used for legacyPackages, devShells, formatter, and checks. It would be better if this tracked unstable rather than a release.
- Unpin the disko flake input and bump to the latest.
- Adds wasgeht package.

## Previous Behavior

n/a

## New Behavior

n/a

## Tests

`nix flake check`
`nix build .#legacyPackages.x86_64-linux.scale-network.wasgeht`

### DIFF
--- a/facts/inventory.py
+++ b/facts/inventory.py
@@ -703,6 +703,11 @@ def generatewasgehtconfig(switches, routers, pis, aps, servers, outputdir):
         wasgehtconfig[ap["name"]] = {"address": ap["ipv4"]}
     for server in servers:
         wasgehtconfig[server["name"]] = {"address": server["ipv6"]}
+    wasgehtconfig["google88v6"] = {"address": "2001:4860:4860::8888"}
+    wasgehtconfig["google88v4"] = {"address": "8.8.8.8"}
+    wasgehtconfig["google44v6"] = {"address": "2001:4860:4860::8844"}
+    wasgehtconfig["google44v4"] = {"address": "8.8.4.4"}
+    wasgehtconfig["localhost"] = {"address": "::1"}
     with open(f"{outputdir}/scale-wasgeht-config.json", "w") as f:
         json.dump(wasgehtconfig, f)
 
--- a/nix/nixos-modules/services/wasgeht.nix
+++ b/nix/nixos-modules/services/wasgeht.nix
@@ -1,0 +1,68 @@
+{
+  config,
+  pkgs,
+  lib,
+  ...
+}:
+let
+
+  cfg = config.scale-network.services.wasgeht;
+
+  inherit (lib)
+    types
+    ;
+
+  inherit (lib.modules)
+    mkIf
+    ;
+
+  inherit (lib.options)
+    mkEnableOption
+    mkOption
+    mkPackageOption
+    ;
+
+in
+
+{
+
+  options.scale-network.services.wasgeht = {
+
+    enable = mkEnableOption "SCaLE wasgeht monitoring service";
+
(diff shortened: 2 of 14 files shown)


djacu:
`core` check is failing

nix/nixos-modules/services/wasgeht.nix
```diff
@@ -0,0 +1,68 @@
...
+      type = types.str;
+      default = "/persist/var/lib/wasgeht";
+    };
+  };
+
+  config = mkIf cfg.enable {
+    systemd.services.wasgeht = {
+      after = [ "network.target" ];
+      serviceConfig = {
+        ExecStart = "${lib.getExe cfg.package} --data-dir=${cfg.statePath} --host-file=${cfg.hostFile} --port=${builtins.toString cfg.port} --log-level={cfg.logLevel}";
```
kylerisse, line 60: - `${lib.getExe cfg.package} is full path to `/nix/store/.../bin/wasgeht` but needs to be `wasgehtd` but and that started it, but other issues.
- `--log-level={cfg.logLevel}` should be `--log-level=${cfg.logLevel}`

nix/nixos-modules/services/wasgeht.nix
```diff
@@ -0,0 +1,68 @@
...
+      };
+      unitConfig = {
+        ConditionPathExists = "!${cfg.statePath}";
```
kylerisse, line 63: this is inverted, we want the path to exist prior to starting.

nix/packages/wasgeht/package.nix
```diff
@@ -0,0 +1,55 @@
...
+    rrdtool
+  ];
+
+  propogatedBuildInputs = [
+    iputils
+    rrdtool
+  ];
+
+  postFixup = ''
+    wrapProgram $out/bin/wasgehtd --set PATH ${
```
sarcasticadmin, line 40: Nice, this makes sense

nix/nixos-modules/services/wasgeht.nix
kylerisse: I've been exploring that after first attempting to go the https://github.com/nix-community/impermanence route and then reversing course. Will continue.

kylerisse: thanks for helping me with this @sarcasticadmin . The permission denied was a red herring. We needed to make wasgeht a `NormalUser` so that it has shell access

facts/inventory.py
```diff
@@ -703,6 +703,11 @@ def generatewasgehtconfig(switches, routers, pis, aps, servers, outputdir):
         wasgehtconfig[ap["name"]] = {"address": ap["ipv4"]}
     for server in servers:
         wasgehtconfig[server["name"]] = {"address": server["ipv6"]}
+    wasgehtconfig["google88v6"] = {"address": "2001:4860:4860::8888"}
+    wasgehtconfig["google88v4"] = {"address": "8.8.8.8"}
+    wasgehtconfig["google44v6"] = {"address": "2001:4860:4860::8844"}
+    wasgehtconfig["google44v4"] = {"address": "8.8.4.4"}
+    wasgehtconfig["localhost"] = {"address": "::1"}
```
kylerisse, line 710: added localhost for the nix checks, also added the upstream Google DNS servers over IPv4 and IPv6 as additional sanity check in the dashboard.

sarcasticadmin:
Closing since it was superseded by #880

Issue #771, opened 2024-09-22
conslidate core, signs, and monitoring vms


sarcasticadmin:
## Description

Rather than running `microvm` well be consolidating all the services to the bare metal on the host. We will still run `kvm` for vms that our running for other teams: AV, etc. 

## Acceptance Criteria

- Consolidate all vms directly to the host
- `vmTests` still pass

PR #846, opened 2025-02-28
nix: finalize consolidation of hypervisor* roles to core*

sarcasticadmin:
## Description of PR

Fixes: #771

`disko` successfully wipe existing hardware for coreMaster using the included config. Additionally, moved dedicated network.nix on coreMaster so it didnt interfere with vmTests. We no longer need the hypervisor* roles so those have been removed. Misc cleanup for `nix.checks` to ensure passing with refactors for core* roles. Removed the last rements of old vm config by deleting ssh4vms.nix and creating ssh.nix (similar config).

## Previous Behavior

- partitions were provisioned by hand
- hypervisor* roles were used to provision the bare metal

## New Behavior

- `disko` for paritioning
- `core`* roles are superseding hypervisor* roles going forward
- New root password hash in time for 22x
- 

## Tests

Installing nixos on existing hardware (utilizing dhcp interface instead of show interface):

```
nix run github:nix-community/nixos-anywhere -- --flake .#coreMaster root@192.168.47.180
```
> We probably want to pull in nixos-anywhere into our flake so its managed instead of running latest

Check on core passes:

```
nix build .#checks.x86_64-linux.core -L
```

Gitlab passing: https://gitlab.com/socallinuxexpo/scale-network/-/jobs/9283798997


sarcasticadmin:
@cjmakes FYI

nix/checks/core.nix
```diff
@@ -75,10 +75,13 @@ in
           inherit inputs;
         };
         imports = [
-          ../nixos-configurations/core-master/configuration.nix
           inputs.self.nixosModules.default
         ];
 
+        scale-network.services.keaMaster.enable = true;
```
sarcasticadmin, line 81: Any objections to something like this? Its easier and less prone to side effects when we need to add specific network config to: `nixos-configurations/core-master/configuration.nix` or similar roles

nix/checks/core.nix
```diff
@@ -87,7 +90,9 @@ in
 
         virtualisation.vlans = [ 1 ];
         systemd.services.systemd-networkd.environment.SYSTEMD_LOG_LEVEL = "debug";
+        networking.useDHCP = false;
         systemd.network = {
+          enable = true;
```
sarcasticadmin, line 95: We need to enable systemd.network if not consuming from `nixos-configurations/core-master/configuration.nix`

nix/nixos-configurations/core-master/configuration.nix
```diff
@@ -27,15 +22,21 @@
...
   };
 
+  systemd.services.systemd-networkd-wait-online.enable = lib.mkForce false;
   # Make sure that the nix/machines/core/master.nixmakes of these files are actually lexicographically before 99-default.link provides by systemd defaults since first match wins
   # Ref: https://github.com/systemd/systemd/issues/9227#issuecomment-395500679
   networking.useDHCP = false;
   systemd.network = {
     enable = true;
+
+    netdevs.virbr0.netdevConfig = {
```
sarcasticadmin, line 32: idea here is to have physical interfaces on a bridge so that if we need to spin up VMs for AV, etc. we can just attach them directly to the bridge for network connectivity. We did pretty much the same thing with hypervisor* last year https://github.com/socallinuxexpo/scale-network/blob/e63ea79dffabafc03b0269b6d59788b301732fed/nix/nixos-configurations/hypervisor1/configuration.nix#L26-L62

nix/nixos-configurations/core-master/default.nix
```diff
@@ -9,13 +9,21 @@
...
+        ./network.nix
+        ./disko.nix
       ];
 
       config = {
         nixpkgs.hostPlatform = "x86_64-linux";
 
         scale-network = {
           base.enable = true;
+          services.keaMaster.enable = true;
```
sarcasticadmin, line 21: now all the scale modules are enabled in the same place

nix/checks/core.nix
cjmakes: only hesitation would be that it makes it easier to drift between prod and test. But not a big concern

cjmakes (approved):

nix/nixos-configurations/core-master/configuration.nix
kylerisse: yup, agreed. this is good solution. I wonder if they'll attempt to setup the `caption` server this year.

nix/checks/core.nix
kylerisse: I object to clustering Kea in the first place, but that ship has sailed. This seems fine though.

nix/nixos-configurations/core-master/disko.nix
```diff
@@ -3,7 +3,7 @@
     disk = {
       one = {
         type = "disk";
-        device = "/dev/disk/by-id/wwn-0x5000c500b794a007";
+        device = "/dev/disk/by-path/pci-0000:10:00.0-scsi-0:0:10:0";
```
kylerisse, line 6: interesting solution. good idea.

nix/checks/core.nix
sarcasticadmin: Clustering Kea isnt actually enabled at the moment, we just run both as `master` and they are independent of one another

nix/nixos-configurations/core-master/configuration.nix
sarcasticadmin: Yes theres nothing definitive but I think we should be ready to accommodate a VM or two.

nix/checks/core.nix
kylerisse: ah my confusion, mixed it up with BIND.

sarcasticadmin: I thought the same thing when I picked this code up from 21x

djacu (approved):
Talked it over with @sarcasticadmin offline about the changes. New module architecture looks good to me.

kylerisse (approved):

cjmakes (approved):

sarcasticadmin:
:shipit: Thanks everyone for the reviews, this is in a much better place after all the feedback :slightly_smiling_face:

PR #850, opened 2025-03-03
initial change to new APs for noc and av [DONE]

davidelang:
change the noc and AV AP definitions to use the new APs

There will be a lot more, but this will enable just these 5 new aps to get started.


kylerisse (approved):
thanks @davidelang

PR #853, opened 2025-03-04
[READY] nixosModules.users.conjones: adds user conjones

cjmakes:
## Description of PR

Adds nix user conjones

## New Behavior

user account conjones exists


sarcasticadmin (approved):
@cjmakes looks good, you should add yourself as enabled to the core* https://github.com/socallinuxexpo/scale-network/blob/205c4364aed957b03bb27e3d15e243a68fa25a18/nix/nixos-configurations/core-master/configuration.nix#L18 for an example

Ill merge after that

PR #856, opened 2025-03-04
assign 125 aps for 2025, set them all on channel 1 and 40 [done]

davidelang:
by setting everything to channels 1 and 40, once they are all up we can issue a scan command to each AP to get a report of what other APs can be heard at what strength.

After this is done, we will deploy a new config that sets the channels to minimize interference (including disabling 2.4GHz)


davidelang:
do not deploy, this will have IP address conflicts

PR #860, opened 2025-03-04
URGENT: Fix errors in switch types file

owendelong:
## Description of PR

Unfucked the switch types file

## Previous Behavior

Generated bogus configurations for trunks in conference center

## New Behavior

Generates correct VLAN maps on conference trunks

## Tests

Broke the NOC
Fixed the NOc
Yak is happy now.
We need to buy more lubricant and surgical tubing.


sarcasticadmin (approved):
ahh yes right building would be helpful

PR #868, opened 2025-03-05
Od 22x patch04

owendelong:
## Description of PR

Fix VLAN and IP address for 2 switches moved from EX to CF

## Previous Behavior

103

## New Behavior

503

## Tests

You put your right yak in,
You put your right yak out,
You give your yak some lube,
and you shake the tube about,
That's what it's all about!


sarcasticadmin (approved):

sarcasticadmin:
phobos :shipit:

PR #869, opened 2025-03-05
facts.inventory: set dhcp4 options to string

sarcasticadmin:
## Description of PR

- Setting this to a string instead of uint8

## Previous Behavior

- dhcp options were uint8

## New Behavior

- dhcp options set to string

## Tests

- Verified working on coreMaster


owendelong (approved):
LGTM

PR #871, opened 2025-03-05
test turning radios off

davidelang:
tests turning radios off, including checking capitalization

after the test is complete we can revert this change


sarcasticadmin:
verified working for #858

PR #881, opened 2025-03-05
split massflash and add state info [WIP] post scale 22

davidelang:
This is not yet tested, being posted for discussion on the approach.

split massflash into two parts:
 1. interface with kea
 2. execute the upgrade with status to stderr

the 2nd part can then be used by other scripts such as
 upgrade subset
 rolling upgrade

### DIFF
--- a/openwrt/scripts/massflash/remoteflash
+++ b/openwrt/scripts/massflash/remoteflash
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+
+export PATH=/run/wrappers/bin:/root/.nix-profile/bin:/etc/profiles/per-user/root/bin:/nix/var/nix/profiles/default/bin:/run/current-system/sw/bin
+
+host="$1"
+mac="$2"
+
+echo "$mac starting, IP $host" >&2
+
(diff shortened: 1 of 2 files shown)


openwrt/scripts/massflash/remoteflash
```diff
@@ -0,0 +1,53 @@
...
+  -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no \
+  $ssh_user@$host "cat /etc/board.json | jq -r .model.id | cut -d',' -f2")
+
+[ -z "ap_type" ] && echo "/etc/board.json type not parsed" && exit 1
+
+echo "$mac AP type $app_type" >&2
+
+scp -O -P $ssh_port \
+  -i $state_dir/id_priv \
+  -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no \
```
kylerisse, line 42: my only suggestion would be to make the ssh args a variable to DRY it over but not worth blocking merge on.

kylerisse (approved):
thanks @sarcasticadmin , glad to see it working and APs flashing success!

kylerisse (changes requested):
let's wait on this @davidelang @sarcasticadmin

PR #882, opened 2025-03-05
[READY] - the stats in /sys do not work for the new APs so find new ones

davidelang:
The prior /sys/kernel/debug* stats do not exist on the openwrt one APs

This version gathers similar data from the new endpoints.

I don't know how this works on the old APs

this just sticks this in syslog


davidelang:
confirmed that the new script works on openWRT one APs

the old script is needed for the ath79 architecture, so move the old script to files-ath79* and make a replacement for the openwrt one

davidelang:
hold, found an interaction that doesn't put the logs in the desired place

PR #886, opened 2025-03-06
Grafana fix

genebean:
## Description of PR

Fix the domain to be the one actually being accessed, add the recommended proxy settings, and add a location block for `/api/live` for Grafana.

## Previous Behavior

Prior to this, Grafana was not working because the domain was set to localhost

## Tests

We don't have a good way to test this right now, but plan to write checks in here once we can validate them.


kylerisse (approved):
thanks @genebean and @BerkhanBerkdemir

PR #893, opened 2025-03-06
Od 22x patch07

davidelang:
rename switches and minor cleanup

testing:
changes are live in the wild


kylerisse (approved):
thanks @owendelong and @davidelang

PR #898, opened 2025-03-06
eliminate dupe name between expo switch and expo ap plugged into that…

davidelang:
… switch


kylerisse (approved):
thanks @davidelang , follow up https://github.com/socallinuxexpo/scale-network/issues/899

PR #894, opened 2025-03-06
nixosModules.service.signs: update signs

sarcasticadmin:
## Description of PR

update signs for latest build for 22x

## Previous Behavior

- scale21x image deployed

## New Behavior

- scale 22x image deployed

## Tests

- http://signs.scale.lan/ is live
- monitoring is up as well on vhost

PR #903, opened 2025-03-06
checks.signs: remove from default since impure

sarcasticadmin:
## Description of PR

relates to: #894

This test is impure,  remove from default.nix for now

## Tests

- tests pass


kylerisse (approved):
got it

PR #908, opened 2025-03-06
Od 22x patch11

owendelong:
## Description of PR

Add DNS resolvers and syslog configuration

## Previous Behavior

No sys logging to log host

## New Behavior

We can has syslog.

## Tests

Kyle sez it works, so it must be good. Mikey has not tasted it yet, but he hates everything except life cereal.


kylerisse (approved):
lgtm @owendelong

PR #911, opened 2025-03-06
Add QoS to destroy the network in lieu of having enough bandwidth

owendelong:
## Description of PR

The physical network in the convention center sucks, so we're making our network suck more in order to not suck even more as everything crashes and burns

## Previous Behavior

APs used to enforce rate limits by nature of their antiquated and limited capabilities.
This prevented us from completely overrunning the PCC network capabilities

## New Behavior

The new APs removed the previous bottlenecks and allowed wifi clients to uttery and completely overrun the network infrastructure.

This patch will allow us to deploy QoS to all non-IDF switches putting rate limits on the most trafficked VLANs and prioritizing AV and Infra traffic.

## Tests

The patch will allow us to convert all of our show attendees into guinnea pigs. It will be fun and exciting and nothing can possibly go wrong.

### DIFF
--- a/switch-configuration/config/scripts/switch_template.pm
+++ b/switch-configuration/config/scripts/switch_template.pm
@@ -1803,6 +1803,8 @@ sub build_config_from_template
   my $hostname = shift @_;
   my $root_auth = shift @_;
   $VV_name_prefix = shift @_;
+  my @switchtype = get_switchtype($hostname);
+  my $type = $switchtype[4];
   
   # Add configuration file fetches here:
   my $TRUNK_MACROS = build_trunktypes_from_config($hostname);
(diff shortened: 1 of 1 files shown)


switch-configuration/config/scripts/switch_template.pm
```diff
@@ -1825,6 +1827,70 @@ sub build_config_from_template
   my $FIREWALL_CONFIG   = ${VENDOR_CONFIGURATION}{"firewall"};
   my $DHCP_CONFIG       = ${VENDOR_CONFIGURATION}{"dhcp"};
   my $PROTOCOL_CONFIG   = ${VENDOR_CONFIGURATION}{"protocols"};
+##FIXME## This is a complete hack to deploy QoS without thinking about it
+##FIXME## QoS should be configurable. Worse, we have to hand-hork this to
+##FIXME## avoid applying it to IDF switches
```
kylerisse, line 1832: so long as we're all clear about what we're doing here

kylerisse (approved):
thanks @owendelong , :shipit:

PR #923, opened 2025-03-07
nixosModules.services.mrtg: init mrtg-generator

sarcasticadmin:
## Description of PR

mrtg-generator service for mrtg

## Tests

- nixos checks core passes
- coreMaster builds as expected


kylerisse (approved):
thank you! live tested.

PR #928, opened 2025-03-07
adds snmp_exporter and scrape jobs to prometheus

cjmakes:
## Description of PR

adds snmp_exporter and scrape jobs to prometheus
## Previous Behavior
no data

## New Behavior
data !
screenshot_2025-03-06-210034

## Tests

I was able to run the same config on my laptop, on the noc vlan and create graphs.

## Notes
The snmp.yaml was generated using https://github.com/prometheus/snmp_exporter/tree/v0.28.0/generator paired down the just the juniper stats we care about

### DIFF
--- a/nix/nixos-modules/services/monitoring.nix
+++ b/nix/nixos-modules/services/monitoring.nix
@@ -66,7 +66,94 @@ in
             }
           ];
         }
+        {
+          job_name = "snmp";
+          static_configs = [
+            {
+              targets = [
+                "noc.scale.lan"
+                "confidf.scale.lan"
+                "nw-idf.scale.lan"
+                "ne-idf.scale.lan"
+                "expoidf.scale.lan"
+                "expo-catwalk.scale.lan"
+                "massflash.scale.lan"
+                "avswitch.scale.lan"
+                "deceased1.scale.lan"
+                "rm211.scale.lan"
+                "rm214.scale.lan"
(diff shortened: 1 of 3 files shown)


sarcasticadmin:
verified working on `coreMaster`

nix/nixos-modules/services/monitoring.nix
```diff
@@ -20,6 +20,13 @@ let
     mkEnableOption
     mkOption
     ;
+
+  unfilteredList = (
+    builtins.split "\n" (
+      builtins.readFile "${pkgs.scale-network.scaleInventory}/config/all-network-devices"
```
kylerisse, line 26: thank you for adding this @cjmakes !

nix/nixos-modules/services/monitoring.nix
```diff
@@ -20,6 +20,13 @@ let
...
+    )
+  );
+  filteredList = (builtins.filter (line: line != [ ] && line != "") unfilteredList);
```
kylerisse, line 29: we might need to make changes to inventory to prevent the need for this.

kylerisse (approved):
Thank you so much @cjmakes . @owendelong will get a lot of value out of this. Thank you also to @BerkhanBerkdemir for getting us to the point of being able to add this type of monitoring!

PR #935, opened 2025-03-07
collect all interface snmp data

cjmakes:
## Description of PR

Collect all snmp data for interfaces, instead of just InOctets

## Previous Behavior
Previously we collected junos specific data and very specific interface stats.

## New Behavior
Do not collect junos specific data, collect all the interface stats.

## Tests
I'm unable to test that this collects and presents the information we want but I've seen this config loaded by the prometheus-snmp-exporter process so it should work.

## Notes
The config needs to be generated using the same release version as the `pkgs.prometheus-snmp-exporter` version. For now this is v0.26.0


sarcasticadmin:
superseded by #927 

Thanks @cjmakes for pushing this up

PR #933, opened 2025-03-07
Include if_mib, system, jnxJsSPUMonitoringMIB

genebean:
These MIBs are the ones we appear to acutally be using.

The Juniper MIB is from https://apps.juniper.net/mib-explorer/navigate?software=Junos%20OS&release=24.4R1&name=jnxJsSPUMonitoringMIB&oid=1.3.6.1.4.1.2636.3.39.1.12.1

PR #939, opened 2025-03-07
Update snmp.yml with SNMP Retries and Timeout Values

MrHamel:
## Description of PR

Add retries and timeouts to Prometheus SNMP exporter.

## Previous Behavior

No retries or timeouts were defined, causing data to not be collected.

## New Behavior

Add retries and timeouts, in the hopes that data will be collected, regardless of how long it takes.

### DIFF
--- a/nix/nixos-modules/services/snmp.yml
+++ b/nix/nixos-modules/services/snmp.yml
@@ -20,6 +20,10 @@ auths:
     version: 2
 modules:
   juniper:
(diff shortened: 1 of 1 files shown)


owendelong (approved):

genebean:
This seems quite reasonable to me, but the live code was cherry picked by @sarcasticadmin from https://github.com/socallinuxexpo/scale-network/pull/933

kylerisse (changes requested):
this is stale and might not align with new monitoring setup, don't merge

PR #942, opened 2025-03-07
Switch QoS order - it appears to be inverted

genebean:
https://www.juniper.net/documentation/us/en/software/junos/cos-security-devices/topics/concept/cos-forwarding-class-overview.html

Based on the info below and on manual testing, it appears that infra is the lowest priority prior right now. This inverts the order so that infra is highest priority.

image


genebean:
For added context, I did a snmp query from coremaster to ballroomg.scale.lan and it took 17s. I ran it again ~ 1 minute later I did the same thing and it took 1m 45s.

owendelong (approved):

PR #943, opened 2025-03-07
More stupid QoS swizzle

owendelong:
QoS -- How to change screwing people over when the network is full from random to deterministic.

Now we're playing with the deterministic values for fun and profit.
(well, maybe just fun.)


MrHamel (approved):
🚀

PR #945, opened 2025-03-07
Added erikreinert and missing gene entry

erikreinert:
## Description of PR

- added new `erikreinert` user
- added missing `gene` entry to core-master


kylerisse (approved):
thanks @erikreinert @genebean

PR #950, opened 2025-03-08
change 5ghz channels to not use 96 or 68

davidelang:
it appears that channels 68 and 96 are either not usable in this area, or not supported by common equipment. No clients have connected to any AP with these channels

so I removed these channels and remapped the others


kylerisse (approved):
thanks @davidelang

PR #954, opened 2025-03-08
Actually Apply QoS to VLANs

MrHamel:
## Description of PR

Actually apply L2 filter to VLANs.

## Previous Behavior

L2 firewall filter was not actually applied to anything.

## New Behavior

L2 firewall filter will be applied to all VLANs.

### DIFF
--- a/switch-configuration/config/scripts/switch_template.pm
+++ b/switch-configuration/config/scripts/switch_template.pm
@@ -950,6 +950,9 @@ sub build_vlans_from_config
     $name {
         description "$desc";
         vlan-id $_;
+	filter {
+            input qos;
+        }
 EOF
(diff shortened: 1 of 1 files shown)


sarcasticadmin (approved):
This looks good :shipit:

sarcasticadmin:
will keep as `[WAIT]` and see if this is needed tomorrow

sarcasticadmin:
@MrHamel did this properly enabled QOS and was this actually applying this to the switches? @owendelong any context you can add?

owendelong:
I think it's moot now... We shouldn't need the QoS with the new routers and even if we do, the switches aren't the place to do it.

sarcasticadmin:
With #1018 this removes the previous QOS configs on the switches so Im going to close this as well

PR #976, opened 2025-09-29
nix: Adds routers skeleton

cjmakes:
## Description of PR

Adds base layout for working on nixos based routers. I went with a structure of each router config lives in a nixos module (`nixos-modules/routers`) and that module can then be included in a nixos-configuration and a nixos test. This pr just contains the absolute basics of creating the modules with multiple interfaces and assigns some ips to those interfaces. Included is also a test that brings up all the routers and ensures they can ping each other.

This should be a good jumping off point for us to develop the more advanced features and config that we will need.

### DIFF
--- a/nix/checks/default.nix
+++ b/nix/checks/default.nix
@@ -56,6 +56,8 @@ genAttrs
     in
     {
       core = pkgs.testers.runNixOSTest (import ./core.nix { inherit inputs lib; });
+      routers = pkgs.testers.runNixOSTest (import ./routers.nix { inherit inputs lib; });
+      router-border = pkgs.testers.runNixOSTest (import ./router-border.nix { inherit inputs lib; });
       loghost = pkgs.testers.runNixOSTest (import ./loghost.nix { inherit inputs; });
       monitor = pkgs.testers.runNixOSTest (import ./monitor.nix { inherit inputs; });
       # impure test and needs to pull container
--- a/nix/checks/router-border.nix
+++ b/nix/checks/router-border.nix
@@ -1,0 +1,35 @@
+{ inputs, lib }:
+{
+  name = "borderrouter";
+
+  nodes = {
+    borderrouter =
+      { lib, ... }:
+      {
+        _module.args = {
+          inherit inputs;
+        };
+        imports = [
+          inputs.self.nixosModules.default
+        ];
+
+        virtualisation.vlans = [
+          1
+          2
+          3
+        ];
+        scale-network = {
+          router.border.enable = true;
+        };
+      };
+  };
+
+  testScript =
+    { nodes, ... }:
+
+    ''
+      start_all()
+      print(borderrouter.succeed("ip -br addr show"))
+    '';
+
+}
--- a/nix/checks/routers.nix
+++ b/nix/checks/routers.nix
@@ -1,0 +1,81 @@
+# This test sets up all the routers and ensures that the network configs are
+# picked up correctly and connectivity works as expected.
+{ inputs, lib }:
+{
+  name = "routers";
+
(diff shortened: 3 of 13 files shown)


MrHamel:
Is it possible to take all the common stuff in the files at `nix/nixos-modules/routers`, and put them into `nix/nixos-modules/default.nix` ? Also, the same for `nix/nixos-configurations/router-*` for `hardware-configuration.nix` ?

This also needs changing from `intel` to `amd`: `hardware.cpu.intel.updateMicrocode = lib.mkDefault config.hardware.enableRedistributableFirmware;`

I'd keep the kernel modules except for `mpt3sas` and `megaraid_sas`, swapping them for `igbe`, `ixgbe`, `i40e`.

For our documentation, the Intel fiber NIC requires Intel coded SFP modules, otherwise the firmware on the card will not accept them, and refuse to provide link. There are ways to get around it (https://www.moduletek.com/en/application_notes/an_00115.html), however patching the firmware is not worth it when getting Intel-coded SFP modules and DAC cables from FS is cheap.

When matching physical interfaces to systemd, let's make sure its documented what `ethX` goes to the onboard NIC, quad copper NIC, and quad fiber NIC. I'd also recommend putting labels on the slots above/below the NICs to know where the cables should go.

What would be nice is if the copper and fiber ports were grouped into columns and configured identically or bridged, meaning....

1 | 2 | 3 | 4 - copper
1 | 2 | 3 | 4 - fiber

(1) ports are border
(2) ports are conference
(3) ports are expo
(4) ports are spare

This would allow us to go 10G by simply moving from the copper port to the fiber port above/below it.

sarcasticadmin:
> Is it possible to take all the common stuff in the files at `nix/nixos-modules/routers`, and put them into `nix/nixos-modules/default.nix` ? Also, the same for `nix/nixos-configurations/router-*` for `hardware-configuration.nix` ?

Yes theres something consolidation that can certainly be done. This is also true for the hardware config for core* machines.

> This also needs changing from `intel` to `amd`: `hardware.cpu.intel.updateMicrocode = lib.mkDefault config.hardware.enableRedistributableFirmware;`
> 
> I'd keep the kernel modules except for `mpt3sas` and `megaraid_sas`, swapping them for `igbe`, `ixgbe`, `i40e`.

To be addressed in #977 

> For our documentation, the Intel fiber NIC requires Intel coded SFP modules, otherwise the firmware on the card will not accept them, and refuse to provide link. There are ways to get around it (https://www.moduletek.com/en/application_notes/an_00115.html), however patching the firmware is not worth it when getting Intel-coded SFP modules and DAC cables from FS is cheap.

Ill check what we already have, but hopefully they are intel SFPs

> When matching physical interfaces to systemd, let's make sure its documented what `ethX` goes to the onboard NIC, quad copper NIC, and quad fiber NIC. I'd also recommend putting labels on the slots above/below the NICs to know where the cables should go.

The intent is to leverage systemd.network.*.matches: https://github.com/socallinuxexpo/scale-network/pull/976#discussion_r2389712986 Whats currently there is just a placeholder for the tests/skeleton. Putting useful interface aliases would be a good idea too now that were talking about about it.

> What would be nice is if the copper and fiber ports were grouped into columns and configured identically or bridged, meaning....
> 
> 1 | 2 | 3 | 4 - copper 1 | 2 | 3 | 4 - fiber
> 
> (1) ports are border (2) ports are conference (3) ports are expo (4) ports are spare
> 
> This would allow us to go 10G by simply moving from the copper port to the fiber port above/below it.

The fiber and copper does not align with this grouping from what I understand. The reason for the cooper ports is because of uplinks to the other IDFs

PR #980, opened 2025-10-16
[READY] nixos-modules: scale-network.services.frr init

djacu:


nix/nixos-modules/services/frr.nix
```diff
@@ -0,0 +1,47 @@
...
+         ip ospf priority 0
+        exit
+        !
+        interface eth2
+         ip address 10.1.3.2/24
+         ip ospf priority 0
+        exit
+        !
+        router ospf
+         network 10.1.1.0/24 area 0
```
kylerisse: `network 10.0.0.0/8 area 0`

kylerisse:
@djacu @sarcasticadmin Pushed up some modifications to OSPF that improve convergence time in `driverInteractive` environment. Also set some passive interfaces and renumbered eth2 on 10.1.3.x to .1 for my own sanity while testing. A current network SME will want to review these values for actual hardware later as they might be too extreme for the live environment.

nix/checks/routers.nix
```diff
@@ -53,29 +75,64 @@
...
       conference.succeed("ping -c 5 10.1.1.1")
       conference.succeed("ping -c 5 10.1.3.3")
 
-      # expo should hit neighbors
+      # expo can ping both routers
       expo.succeed("ping -c 5 10.1.2.1")
       expo.succeed("ping -c 5 10.1.3.2")
+
+      # layer 3
+      print("LAYER 3")
```
sarcasticadmin: @djacu and I added some checks here for frr and the routes that are learned by each router via ospf

nix/nixos-modules/services/frr.nix
```diff
@@ -0,0 +1,111 @@
...
+    mkEnableOption
+    mkOption
+    ;
+
+  inherit (lib.strings)
+    concatStringsSep
+    ;
+in
+{
+  options.scale-network.services.frr = {
```
sarcasticadmin: We consolidated our config into one module instead of `frrN.nix`

sarcasticadmin (commented):
@kylerisse @cjmakes @erikreinert Weve got the vmtest now passing across all 3 routers with a single frr module. Lets us know what you think. I think we can get feedback and if all looks good we can fixup the commits then merge as its a a good point in time.

nix/checks/routers.nix
```diff
@@ -53,29 +75,64 @@
...
+      print("LAYER 2")
+
+      # border can ping both routers
       border.succeed("ping -c 5 10.1.1.2")
       border.succeed("ping -c 5 10.1.2.3")
 
-      # conference should hit neighbors
+      # conference can ping both routers
       conference.succeed("ping -c 5 10.1.1.1")
       conference.succeed("ping -c 5 10.1.3.3")
```
kylerisse: can this be 10.1.3.1? It's mostly a sanity thing when testing (each link has a .1 and .2)

sarcasticadmin: Thats fine, its just currently set that expo uses .3 for the last octet on all of interfaces: https://github.com/socallinuxexpo/scale-network/blob/f505f8b178d79a2461de3d81a0b2714542e45dca/nix/nixos-modules/routers/expo.nix#L39 so we'd need to update it there as well

sarcasticadmin:
Thank you @djacu @erikreinert @kylerisse @cjmakes for all the help with this, great progress! :rocket:

PR #981, opened 2025-10-16
treewide: use mapAttrs on a flake output over genAttrs everywhere

djacu:
The cleans up re-defining list of systems everywhere.

_Should_ be a no-op.


sarcasticadmin (approved):
paired with @djacu on this, looks good :rocket:

sarcasticadmin:
Thanks for the clean up @djacu !

PR #982, opened 2025-10-18
[READY] simplify module imports

djacu:
- **nixos-modules: simplify imports**
- **nixos-modules: move bind-master**
- **nixos-modules: remove bhyve-image**


djacu:
> @djacu lets remove `bhyve-image.nix` we havent needed it for several years

Done

PR #983, opened 2025-10-23
[READY] - Implement LGTM on core-master

erikreinert:
## Changes

- added LGTM stack to existing monitoring configuration

### DIFF
--- a/nix/checks/monitor.nix
+++ b/nix/checks/monitor.nix
@@ -11,7 +11,10 @@
       inputs.self.nixosModules.default
     ];
     virtualisation.graphics = true;
-    scale-network.services.monitoring.enable = true;
+    scale-network.services = {
+      monitoring.enable = true;
+      alloy.enable = true;
+    };
   };
 
   nodes.client1 =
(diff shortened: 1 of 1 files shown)


nix/checks/monitor.nix
```diff
@@ -14,29 +13,27 @@
...
-  testScript =
-    { nodes, ... }:
-    ''
-      start_all()
-      coremaster.succeed("sleep 2")
-      coremaster.wait_for_unit("grafana.service", None, 30)
-      coremaster.wait_for_unit("prometheus.service", None, 30)
-      coremaster.wait_until_succeeds("nc -vz localhost 3000")
+    client1.wait_until_succeeds("ping -c 5 ${nodes.coremaster.networking.hostName}")
+    client1.wait_until_succeeds("curl -v -L -H \"Host: monitoring.scale.lan\" http://${nodes.coremaster.networking.hostName}")
```
jshcmpbll: From what I have found, NixOS VM testing overrides the hostname with the node name defined in the test config and unfortunately you can not have `.` in the name. So in this case you would have to replace the first one with:

```
client1.wait_until_succeeds("ping -c 5 coremaster")
```

And the second check won't work since the Nginx proxy was removed.

jshcmpbll (commented):
Hi @erikreinert  - Rob had asked me to take a look at this PR during the SCALE work party today in Burbank. I tried to add some helpful comments but open to discussing this further if you are interested or need help with anything. I'm pretty familiar with Grafana and some of the other services but I haven't stayed up to date with some of the newer ones like Tempo or Mimir.

Also, @kylerisse had mentioned he had a good solution for setting up SSL certificates so if we wanted to bring the Nginx proxy back in we could establish a secure connection to the Grafana web server.

PR #985, opened 2025-11-06
nixosConfig.router-*: init the new hardware

sarcasticadmin:
## Description of PR

Depends on: https://github.com/socallinuxexpo/scale-network/pull/980
Fixes: https://github.com/socallinuxexpo/scale-network/issues/977

## Previous Behavior

- Router hardware didnt have a config

## New Behavior

- `router-border`, `router-conf`, `router-expo` have hardware config

## Tests

- Tested against our 3 physical routers

Issue #977, opened 2025-09-30
Declare new router hardware specs for network and disk in nix


sarcasticadmin:
## Description

Follow up to: https://github.com/socallinuxexpo/scale-network/pull/976

With the creation of the new routers we have to define the network cards and disk inside of nix.

## AC

- Update systemd.network for the appropriate cards/ports/driver matches
- Update the disko config for specific PCI lane info for NVME cards
- Confirm nixos-anywhere works with entire nixos-configuration for each router

PR #987, opened 2025-11-07
[READY] nixosConfiguratios.router-*: NAT, frr, static route

sarcasticadmin:
## Description of PR

Relates to: https://github.com/socallinuxexpo/scale-network/issues/977
Follow up to: https://github.com/socallinuxexpo/scale-network/pull/985

Reverted the commit in question and setting up NAT on for nixosConfigurations.router-border. Also setting DHCP on copper0 for NAT to make testing easy.

## Previous Behavior

- NAT nftables on router-expo (incorrect)
- copper0 had no network config
- firewall on router-{expo, conf}
- no static router to border 

## New Behavior

- NAT nftables on router-border
- copper0 DHCP
- disabled firewall on router-{expo,conf}
- static routes for to border
- remove extra ospf config timers and fast reroute

## Tests
- confirm working on all router hardware

### DIFF
--- a/nix/nixos-configurations/router-expo/default.nix
+++ b/nix/nixos-configurations/router-expo/default.nix
@@ -44,6 +44,8 @@
           SUBSYSTEM=="net", ACTION=="add", KERNELS=="0000:0d:00.3", NAME="copper3"
         '';
 
+        networking.firewall.enable = false;
+
         # must be disabled if using systemd.network
         networking.useDHCP = false;
 


kylerisse (approved):
nice catch @MrHamel @sarcasticadmin .

MrHamel:
If `copper0` is the WAN uplink to $ISP, it must be configurable with static IPs. It's okay to use DHCP outside the convention center for various testing.

Something to keep in mind is the "chicken and egg" problem, in that "how do I get the border router bootstrapped with the static IP(s) without an Internet connection at the convention center?"

MrHamel:
Also need to disable OSPF on all unnecessary ports. The below config disables it globally, and we enable it as required. OSPF was sending hello packets out the ISP uplink, which is a big no-no, especially with no route-map dropping unknown routes.

router ospf
 passive-interface default

int fiber0
 no ip ospf passive

int fiber1
 no ip ospf passive
________________________________

PR #990, opened 2025-11-09
Massflash fixes

jfly:
This fixes a few issues discovered while attempting to use massflash today. See the individual commits for details.


sarcasticadmin (approved):
@jfly @MrHamel @djacu :rocket: thanks for bring massflash back from the dead

PR #994, opened 2025-11-09
Port massflash to use `writeShellApplication`

jfly:
See the individual commits for details.

### DIFF
--- a/nix/packages/massflash/massflash
+++ b/nix/packages/massflash/massflash
@@ -7,13 +7,13 @@ hook=$1
 case $hook in
   leases4_committed)
     echo "lease4_committed"
-    echo $QUERY4_REMOTE_ADDR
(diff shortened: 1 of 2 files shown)


nix/packages/massflash/massflash
```diff
@@ -7,13 +7,13 @@ hook=$1
 case $hook in
```
sarcasticadmin: shellcheck :heart:

nix/packages/massflash/package.nix
```diff
@@ -1,28 +1,13 @@
...
   openssh,
   bash,
 }:
-let
-  local_manifests = copyPathsToStore [
-    ../../../openwrt/scripts/massflash/massflash
-  ];
-in
-stdenvNoCC.mkDerivation {
+writeShellApplication {
```
sarcasticadmin: so much better!

PR #995, opened 2025-11-09
[READY] treewide: refactor overlay structure

djacu:
## Description of PR

- Add nixpkgs-lib to provide the nixpkgs library rather than nixpkgs-unstable.
- Extend the library output to add new path functions.
- Move all local package additions to `package-sets`.
- Use `packagesFromDirectoryRecursive` to create the initial top-level overlay.
- Create a new `package-overrides` overlay directory and move the frr overrides into it.

## Previous Behavior

Packages names were converted to camel case from their kebab case directory.

## New Behavior

:camel: :gun: 

## Tests

What?

### DIFF
--- a/nix/package-sets/top-level/scale-network/scale-inventory/package.nix
+++ b/nix/package-sets/top-level/scale-network/scale-inventory/package.nix
@@ -7,8 +7,8 @@
 }:
 let
   local_manifests = copyPathsToStore [
-    ../../../switch-configuration
-    ../../../facts
+    ../../../../../switch-configuration
+    ../../../../../facts
   ];
 in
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -5,6 +5,7 @@
     disko.inputs.nixpkgs.follows = "nixpkgs-unstable";
     disko.url = "github:nix-community/disko";
     nixpkgs-2505.url = "github:NixOS/nixpkgs";
+    nixpkgs-lib.url = "github:nix-community/nixpkgs.lib";
```
sarcasticadmin: TIL

PR #1020, opened 2026-01-08
[READY] nixos-modules.routers.*: vlans and bridges for ospf area0

sarcasticadmin:
## Description of PR

vlans and bridges for the ospf network to match previous config from junipers SRXs

## Tests

- vm tests pass: `nix run .#checks.x86_64-linux.routers.driver`
- Applied to existing bare metal config and passes after restarting frr on each node:

```
# systemctl stop frr.service

[root@router-expo:~]# systemctl start frr.service

[root@router-expo:~]# vtysh

Hello, this is FRRouting (version 10.4.1).
Copyright 1996-2005 Kunihiro Ishiguro, et al.

router-expo# show ip ospf neighbor

Neighbor ID     Pri State           Up Time         Dead Time Address         Interface                        RXmtL RqstL DBsmL
10.1.1.1          1 Full/DR         17.260s            2.718s 10.1.2.1        bridge901:10.1.2.3                   0     0     0
10.1.1.2          1 Full/DR         17.261s            2.256s 10.1.3.2        bridge902:10.1.3.3                   0     0     0

router-expo#
````

### DIFF
--- a/nix/nixos-modules/routers/border.nix
+++ b/nix/nixos-modules/routers/border.nix
@@ -63,19 +63,76 @@ in
     systemd.network = {
       enable = true;
 
+      netdevs = {
+        "20-bridge900" = {
+          netdevConfig = {
+            Kind = "bridge";
+            Name = "bridge900";
+          };
+        };
+        "20-vlan900" = {
+          netdevConfig = {
+            Kind = "vlan";
+            Name = "vlan900";
+          };
+          vlanConfig.Id = 900;
+        };
+        "25-bridge901" = {
+          netdevConfig = {
+            Kind = "bridge";
+            Name = "bridge901";
+          };
+        };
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: nix/nixos-modules/routers/border.nix [hunk 1]
```diff
@@ -135,8 +192,8 @@ in
       services.frr.enable = true;
       services.frr.router-id = "10.1.1.1";
       services.frr.broadcast-interface = [
-        cfg.frrExpoInterface
-        cfg.frrConferenceInterface
+        "bridge900" # cf
+        "bridge901" # expo
```
kylerisse: I like these comments, nice compromise if not using a `cfg` value. possible to add to all 3 so it's clear later which inteface goes which way?

PR #1041, opened 2026-01-31
*&^%*& 256==0 -- Only in BSD.

owendelong:
## Description of PR

Turns out that BSD can't tell 256 from 0

## Previous Behavior

256 == success

## New Behavior

255 != success

## Tests

Nah.


switch-configuration/config/scripts/switch_initial_loader.sh
```diff
@@ -47,7 +47,7 @@ result=$?
 if [ "$result" -eq 0 ]; then
   echo "Configuration CGI returned error:"
   cat /tmp/config.txt | sed -e 's/^/	error: /'
-  exit 256
+  exit 255
```
sarcasticadmin: TIL

sarcasticadmin (approved):

PR #1049, opened 2026-02-01
[READY] add initial ap location data

davidelang:
partially populate the location data for APs for mapping purposes

currently this data is not used

Thanks to Heather for preparing the maps and Lance for doing the data entry

This also includes the tool and instructions for populating map data into a csv


davidelang:
@kylerisse for the x/y locations, when we first created these fields, they had no useful data in them (and initially we had been thinking in terms of pixel counts), but going forward they will either be a percentage or measured in ft, either way a float or in (i.e. allow 50 rather than require 50.0) is correct

davidelang:
removed software from pr

davidelang:
reopened

PR #1058, opened 2026-02-01
router*: ensure ixgbe kernel module is loaded on bootup

jmbaur:
## Description of PR

Currently ixgbe must be manually loaded, adding it to `boot.kernelModules` ensures it is loaded when the machine boots up (post stage 1).

## Previous Behavior

We had the module parameter `allow_unsupported_sfp` successfully configured, but were not loading the module in an automated manner.

## New Behavior

The module is now both correctly configured as well as automatically loaded.

## Tests

On a freshly booted machine with this config:
```
$ lsmod  | grep ixgbe
ixgbe                 487424  0
...
$ cat /sys/module/ixgbe/parameters/allow_unsupported_sfp 
Y
```

### DIFF
--- a/nix/nixos-configurations/router-border/default.nix
+++ b/nix/nixos-configurations/router-border/default.nix
@@ -20,6 +20,7 @@
         boot.extraModprobeConfig = ''
           options ixgbe allow_unsupported_sfp=1,1
         '';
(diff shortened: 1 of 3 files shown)


sarcasticadmin:
Awesome thanks man 🚀

PR #1062, opened 2026-02-01
[READY] Add SNMP scraping for network switches

erikreinert:
## Summary

- Add "type" field to all device data structures (vlan, switch, router, ap, pi, server) to enable future type-based filtering and processing
- Generate `prom-switches.json` targeting port 161 for SNMP polling of Juniper switches
- Configure Alloy with SNMP exporter to collect switch metrics (port statistics, CPU, memory) alongside existing AP and Pi metrics

### DIFF
--- a/nix/nixos-modules/services/monitoring.nix
+++ b/nix/nixos-modules/services/monitoring.nix
@@ -292,7 +292,7 @@ in
 
         // SNMP exporter for Juniper switches
         prometheus.exporter.snmp "switches" {
-          config_file = "${../snmp.yml}"
+          config_file = "${./snmp.yml}"
           targets = discovery.file.switches.targets
         }
 


facts/inventory.py
```diff
@@ -930,6 +936,17 @@ def generatepromconfigs(pis, aps, outputdir):
     with open(f"{outputdir}/prom-pis.json", "w") as f:
         f.write(json.dumps(prom_pi_config, indent=2))
 
+    prom_switch_config = [
```
sarcasticadmin: seems reasonable, thanks for sticking with the existing convention

facts/test_inventory.py
```diff
@@ -488,3 +507,51 @@ def test_populateservers():
     ]
     for filename, servers in cases:
         assert inventory.populateservers(filename, mocvlans) == servers, filename
+
+
+def test_generatepromconfigs():
```
sarcasticadmin: i dont think these tests add a lot of value, seems to be mostly testing that `open` works. Lets remove these tests entirely

nix/nixos-modules/services/monitoring.nix
```diff
@@ -284,6 +284,26 @@ in
...
+        }
+
+        // SNMP exporter for Juniper switches
+        prometheus.exporter.snmp "switches" {
+          config_file = "${./snmp.yml}"
+          targets = discovery.file.switches.targets
+        }
+
+        // Scrape SNMP metrics from switches
+        prometheus.scrape "switches" {
```
sarcasticadmin: nice

sarcasticadmin (approved):
Updated this PR with the subset of the original changes from `rh/snmp-scrape`

PR #1064, opened 2026-02-02
nix: add nix-based AP configuration

jmbaur:
## Description of PR

This builds on top of [mixos](https://github.com/jmbaur/mixos) and [openwrt-one.nix](https://github.com/jmbaur/openwrt-one.nix) in order to use nix instead of openwrt for building the scale APs. There are still parts of this that don't work (notably 5GHz WiFi), but a good amount of this is ready to go.

## Previous Behavior

N/A

## New Behavior

N/A


nix/mixos-configurations/ap/default.nix
```diff
@@ -0,0 +1,192 @@
...
+  };
+
+  init.hostapd = {
+    action = "respawn";
+    process = "${lib.getExe' pkgs.hostapd "hostapd"} ${./hostapd-scaleslow.conf}";
+  };
+
+  init.udhcpc = {
+    action = "respawn";
+    process = "/bin/udhcpc -f -S -i mgmt-br -O 224 -O 225 -O 226 -s ${pkgs.writeScript "udhcpc-script.sh" ''
```
sarcasticadmin: clean, I was worried how much of a pain this might be to mimic from openwrt

.gitignore
```diff
@@ -20,7 +20,7 @@ switch-configuration/config/output
 switch-configuration/config/switch-maps
 
 # Ignore nix-isms
-result
+result*
```
sarcasticadmin: thanks for catching this

flake.nix
```diff
@@ -19,6 +24,7 @@
     formatterModule = import ./nix/formatterModule inputs;
     legacyPackages = import ./nix/legacy-packages inputs;
     library = import ./nix/library inputs;
+    mixosConfigurations = import ./nix/mixos-configurations inputs;
```
sarcasticadmin: `mixosConfigurations` love it

nix/mixos-configurations/ap/hostapd-scaleslow.conf
```diff
@@ -0,0 +1,32 @@
+channel=0
```
sarcasticadmin: hows this compare to what ships in the openwrt image? Just curious since I never dumped hostapd

flake.nix
jmbaur: :sweat_smile: we just need `nix flake show` to be happy now

nix/mixos-configurations/ap/default.nix
```diff
@@ -0,0 +1,192 @@
...
+      #!/bin/sh
+      ${pkgs.busybox}/default.script "$@"
+      ${lib.fileContents ./udhcpc-script.sh}
+    ''}";
+  };
+
+  init.create-interfaces = {
+    action = "wait";
+    process = pkgs.writeScript "create-interfaces.ash" ''
```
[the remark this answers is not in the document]
jmbaur: Just ash! The default system really just comes with busybox and a mixos binary that facilitates some custom system initialization steps. That's not to say bash can't be added, and it really wouldn't add much bloat to the system. We've got all the conveniences of nixpkgs here, so `writeShellApplication` is totally possible. I tend to use `writeScript` with a `/bin/sh` shebang line for mixos stuff since it does prevent unintentional closure bloat.

nix/mixos-configurations/ap/hostapd-scaleslow.conf
sarcasticadmin: Nice ACS makes sense on initial bring up

And agreed, writing to the config template for reboot persistence is a great idea

flake.lock
```diff
@@ -52,11 +72,11 @@
     },
     "nixpkgs-unstable": {
       "locked": {
-        "lastModified": 1756542300,
```
sarcasticadmin: I didnt notice this earlier but do we need to bump this for whats need for mix? 2 things Im concerned about with a bump: Currently this will effect all nixosConfigurations. We did intend to use 25.11 for all the configs but just hadnt cut over to that yet.

One of the perl modules needs a patch due to GCC 15 failure: `$ nix build .#perl-net-interface` Debian also already noted this failure and a patch: https://rt.cpan.org/Public/Bug/Display.html?id=169596

Issue #1006, opened 2025-11-20
linux routers need to enable DHCP helper configs


sarcasticadmin:
## Description

The previous routers provided DHCP helper functionality for various vlans on the network. We need to make sure the linux based routers are capable of doing the same.

## AC

- Linux routers support DHCP helper functionality across necessary VLANs

PR #1101, opened 2026-02-22
[READY] nixos-modules.routers.{conf,expo}: ipv6 relay config

sarcasticadmin:
## Description of PR

depends on: #1098 
fixes: #1006

Remaining relay config for conf and expo routers

## Previous Behavior
- outdated option config for dhcp6relay service
- no dhcp6relay on conf or expo

## New Behavior
- update option names for dhcp6relay service to match dhcp4relay
- enable dhcp6relay on conf or expo

## Tests
- Tested against bare metal conf router
- Tested against bare metal expo router (it applies but there is nothing on copper0 to connect to)

### DIFF
--- a/nix/nixos-modules/routers/expo.nix
+++ b/nix/nixos-modules/routers/expo.nix
@@ -399,19 +399,40 @@ in
         "bridge903" # conf
       ];
 
-      services.dhcp4-relay."tech" = {
-        enable = true;
-        # excluding bridge107 (exSigns) since
-        # its a ipv6 only network
-        downstreamInterfaces = [
-          "bridge100"
-          "bridge101"
-          "bridge102"
-          "bridge104"
-          "bridge110"
-        ];
-        upstreamInterfaces = [ "bridge103" ];
-        dhcpServerIps = [ "10.0.3.20" ];
+      services.dhcp4-relay = {
+        # no AV dhcp6-relay because dhcp server on the same subnet (105)
+        # as its clients
+        "tech" = {
+          enable = true;
+          # excluding bridge107 (exSigns) since
+          # its a ipv6 only network
+          downstreamInterfaces = [
+            "bridge100"
+            "bridge101"
+            "bridge102"
+            "bridge104"
+            "bridge110"
+          ];
+          upstreamInterfaces = [ "bridge103" ];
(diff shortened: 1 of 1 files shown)


nix/nixos-modules/services/dhcp6-relay.nix
```diff
@@ -70,8 +70,8 @@ in
 
         serviceConfig = {
           ExecStart = "${pkgs.scale-network.isc-dhcp}/bin/dhcrelay -6 -d --no-pid ${
-            concatMapStringsSep " " (x: "-l ${x}") relayCfg.listenInterfaces
-          } ${concatMapStringsSep " " (x: "-u ${x}") relayCfg.dhcpServerIps}";
+            concatMapStringsSep " " (x: "-l ${x}") relayCfg.downstreamInterfaces
+          } ${concatMapStringsSep " " (x: "-u ${x}") relayCfg.upstreamInterfaces}";
```
kylerisse: Pretty cool, works just like ipv4 version and easy to follow. built toplevel for `router-conf` and verified unit files for `dhcp6-relay-tech.service` and `dhcp6-relay-av.service` rendered. It's kind of weird that the end of the `ExecStart` line for v6 is formatted like this `-u 2001:470:f026:503::20%%bridge503` with the percents and interface names but for v4 is just the IP. Is this some IPv6 specific thing?

kylerisse (commented):
@sarcasticadmin I had one question about how the unit file for v6 is rendered but if you tested on bare metal then it probably makes sense. just more of a curiosity.

Also notice no `result/etc/systemd/system/dhcp4-relay-av.service` yet? another pass or should that be in this one also?

Thanks for getting rid of that top level `monitoring` folder. That stuff never went anywhere. Excited from what I've seen so far with @erikreinert 's contributions!

sarcasticadmin:
> @sarcasticadmin I had one question about how the unit file for v6 is rendered but if you tested on bare metal then it probably makes sense. just more of a curiosity.
> 
> Also notice no `result/etc/systemd/system/dhcp4-relay-av.service` yet? another pass or should that be in this one also?


100%, I should have noted that I was still WIP on this but wanted to get review on what I had so far

PR #1104, opened 2026-02-27
[READY] - nixos-modules.routers.*: frr updates and ipv6 tests

sarcasticadmin:
## Description of PR

Including the ospf6 daemon and appropriate v6 links. Additionally support trunk interfaces as args to routers to be able to test between vm and bare metal. Thank you @kylerisse and @djacu for helping with this!

## Previous Behavior
- ipv4 routes on router ospf vlans
- ospf daemon only
- cooper0 assumed trunk interface

## New Behavior
- ipv6 routes added to vlan90x vlans
- ospf and ospf6 services running
- added trunk Interfaces option for routers.{conf, expo} 
- vm tests for ipv6 links
- vm tests for relays and radvd

## Tests
- Tests pass in CI
- Tests pass on bare metal
```
[root@router-conf:~]# vtysh -c 'show ipv6 route'
Codes: K - kernel route, C - connected, L - local, S - static,
       R - RIPng, O - OSPFv3, I - IS-IS, B - BGP, N - NHRP,
       T - Table, v - VNC, V - VNC-Direct, A - Babel, D - SHARP,
       F - PBR, f - OpenFabric, t - Table-Direct,
       > - selected route, * - FIB route, q - queued, r - rejected, b - backup
       t - trapped, o - offload failure

IPv6 unicast VRF default:
C>* 2001:470:f026:500::/64 is directly connected, bridge500, weight 1, 00:09:20
K * 2001:470:f026:500::/64 [0/256] is directly connected, bridge500, weight 1, 00:09:22
L>* 2001:470:f026:500::1/128 is directly connected, bridge500, weight 1, 00:09:20
C>* 2001:470:f026:501::/64 is directly connected, bridge501, weight 1, 00:09:20
K * 2001:470:f026:501::/64 [0/256] is directly connected, bridge501, weight 1, 00:09:22
L>* 2001:470:f026:501::1/128 is directly connected, bridge501, weight 1, 00:09:20
C>* 2001:470:f026:502::/64 is directly connected, bridge502, weight 1, 00:09:20
K * 2001:470:f026:502::/64 [0/256] is directly connected, bridge502, weight 1, 00:09:22
L>* 2001:470:f026:502::1/128 is directly connected, bridge502, weight 1, 00:09:20
C>* 2001:470:f026:503::/64 is directly connected, bridge503, weight 1, 00:09:20
K * 2001:470:f026:503::/64 [0/256] is directly connected, bridge503, weight 1, 00:09:22
[...]

PR #1105, opened 2026-02-27
[READY] nixos: repo sync and disable gw on backdoor0

sarcasticadmin:
## Description of PR

Depends on: https://github.com/socallinuxexpo/scale-network/pull/1104

Misc items that we wanted to include on the routers

## Previous Behavior
- No local copy of the repo which makes rebuild-switch trick if internet is down
- backdoor0 received gw from dhcp
- `carrier` was set on trunk interfaces which meant if not all were connected then systemd.network-online couldnt be reached

## New Behavior
- Include syncRepo script to ensure we have a update to date copy of scale-network. This is in `base.nix` so its added to all machines (`core-*`, `dev-server`, etc.)
- disable gw on backdoor0 interface so it doesnt accidently interfere with ospf
- set `no-carrier` on trunk interfaces so we are not waiting for them to reach systemd.network-online state

## Tests
- Applied to bare metal conf router

### DIFF
--- a/nix/nixos-modules/routers/conference.nix
+++ b/nix/nixos-modules/routers/conference.nix
@@ -230,7 +230,7 @@ in
             nameValuePair "30-${interface}" {
               matchConfig.Name = "${interface}";
               linkConfig = {
-                RequiredForOnline = "carrier";
+                RequiredForOnline = "no-carrier";
               };
               networkConfig = {
                 LinkLocalAddressing = "no";
--- a/nix/nixos-modules/routers/expo.nix
+++ b/nix/nixos-modules/routers/expo.nix
@@ -226,7 +226,7 @@ in
             nameValuePair "30-${interface}" {
               matchConfig.Name = "${interface}";
               linkConfig = {
-                RequiredForOnline = "carrier";
+                RequiredForOnline = "no-carrier";
               };
               networkConfig = {
                 LinkLocalAddressing = "no";


nix/nixos-modules/base.nix
```diff
@@ -22,6 +22,27 @@ in
...
+
+        if [ ! -d "$REPO_DIR" ]; then
+            echo "Cloning repository..."
+            ${lib.getExe pkgs.git} clone "$REPO_URL" "$REPO_DIR"
+        else
+            echo "Repository exists. Fetching updates..."
+            ${lib.getExe pkgs.git} -C "$REPO_DIR" fetch --all --prune
+        fi
+      '';
+    };
```
kylerisse: this is cool, nice add! also thanks for taking into account the need for fine grained GH token in our covno.

kylerisse (approved):
lgtm @sarcasticadmin !

nix/nixos-modules/routers/conference.nix
```diff
@@ -209,191 +223,198 @@ in
...
-          networkConfig = {
-            Bridge = "bridge903";
+      networks =
+        let
+          trunks =
+            interface:
+            nameValuePair "30-${interface}" {
+              matchConfig.Name = "${interface}";
+              linkConfig = {
+                RequiredForOnline = "no-carrier";
```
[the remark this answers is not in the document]
sarcasticadmin: just to clarify: the physical interface link status is ignored for `systemd.network-online` to be considered online

nix/nixos-modules/base.nix
sarcasticadmin: Yep Ill be add that to each of our boxes this weekend

Issue #1106, opened 2026-02-27
Improve sticker generation for Juniper toy switches


owendelong:
### Description

Currently requires magic naming to make sticker the correct size cut line

### AC

Improve the sticker generation process so that the sticker generator can detect switch type without requiring magic name of switch.

PR #1111, opened 2026-03-01
[READY] Better sticker size detection

owendelong:
- switchtypes renamed switches away from magic naming for toy switches
	- switch_template -- Many mods to postscript to accomodate tiny labels
	- generate_ps_stickers -- Improved detection of switch Model Side effect, titles for labels now include switch model as well as type

## Description of PR

Fixes: #1106

Improved code for handling toy switches

## Previous Behavior

Magic switch naming convention to trigger smaller cut lines for stickers to go on toy switches
Most (all) titles too wide to fit properly on toy switch labels

## New Behavior

No more magic naming convention, now based on switch model (default full size)
Added a place in the code that needs to be updated if we add more switches (not ideal).
Use smaller title font if Title doesn't fit (this will be applied to a full size switch if needed, too)

Currently, we don't test for max width on switches with fiber ports, so some titles are truncated behind fiber port displays on the few switches that have fiber ports configured.

## Tests
make
open config/switch-maps/stickers.pdf
Use mark 1 eyeball (x2) to visually inspect results

### DIFF
--- a/switch-configuration/config/scripts/switch_template.pm
+++ b/switch-configuration/config/scripts/switch_template.pm
@@ -11,7 +11,6 @@
 ##FIXME## Build a consistency check to match up VLANs in the vlans file(s) and
 ##FIXME## those defined in the types/* files.
 
-##FIXME## Learn v6 prefix dynamically from config files (look for 2001:470 throughout script)
 ##FIXME## Assumes v6 prefix is a /48 and many code dependencies on this assumption. Safe at least for now.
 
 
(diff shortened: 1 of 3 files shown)


switch-configuration/config/scripts/switch_template.pm
```diff
@@ -1949,12 +1958,12 @@ system {
...
     ntp {
-        server 2001:470:f026:503::20;
-        server 2001:470:f026:103::20;
+        server $PREFIX:503::20;
+        server $PREFIX:103::20;
     }
     name-server {
-        2001:470:f026:503::20;
-        2001:470:f026:103::20;
+        $PREFIX:503::20;
```
sarcasticadmin: thanks for parameterizing the ipv6 prefix

PR #1130, opened 2026-03-04
[READY] - MixOS AP updates

jmbaur:
## Description of PR

Updates openwrt-one.nix and provides a small fix for not all links being up prior to wireless interfaces being attached to bridges.

## Previous Behavior

The bridges that the wireless interfaces were attached to were not set up on boot, meaning there was no end-to-end connectivity through the AP.

## New Behavior

Bridges are set up early on in the boot cycle, meaning that as soon as the wireless interfaces are attached to their bridge, wireless clients can use the AP.

## Tests

Tested end-to-end connectivity with two machines, one connected to an RJ45 port on the AP with a VLAN ID 100 interface configured and another connected to the AP via the 2.4Ghz radio.


nix/mixos-configurations/ap/hostapd-scaleslow.conf
```diff
@@ -22,7 +22,7 @@ logger_syslog_level=2
 macaddr_acl=0
 rsn_pairwise=CCMP
 sae_require_mfp=1
-ssid=scaleslow
```
jmbaur: This was done to create a clear distinction between the openwrt-based APs and these. LMK if we want this to look different.

sarcasticadmin (approved):
Time to go live

PR #1134, opened 2026-03-04
Another late change request to the room list

owendelong:
## Description of PR

Rm207 added after deadline 

## Previous Behavior

Rm207 wasn't expected to be a room or need a switch.

## New Behavior

Rm207 is configured against what was SpareG (fortunately the label will match except for the name)

## Tests

Because it's late to the party, it's completely untested, but it should probably work ,maybe.


kylerisse (approved):
lgtm

PR #1165, opened 2026-03-06
Bump openwrt-one.nix

jmbaur:
This brings in updates to the U-Boot firmware that adds support for GPT partition tables, which allows for reading from USB sticks plugged into the device.


sarcasticadmin (approved):
Nice!

Issue #964, opened 2025-03-24
Make sudo on all nix-based systems use ssh keys for auth


owendelong:
Found this and thought it would be a better way to handle sudo than handing out the root password:

https://www.vinnie.work/blog/2023-01-28-sudo-with-ssh-keys

owendelong:
This turned out to be super easy because the capability is already there in an existing NIX module and just needs to be turned on.

owendelong:
Should be closed as completed with merge of #1187

sarcasticadmin:
closed via https://github.com/socallinuxexpo/scale-network/pull/1187

PR #1187, opened 2026-03-31
[READY] Enable ssh-key authentication for sudo via rssh

owendelong:
## Description of PR

See #964 

## Previous Behavior

sudo used root password.

## New Behavior

sudo is passwordless as long as the user's SSH key is available from their SSH Agent or agent forwarding.

## Tests

nixos-rebuild with this branch on both scale-devserver and br-fmt2

### DIFF
--- a/nix/nixos-modules/users/root.nix
+++ b/nix/nixos-modules/users/root.nix
@@ -23,7 +23,6 @@ in
 
     security.sudo = {
       extraConfig = ''
-#        Defaults rootpw
         Defaults lecture="never"
       '';
     };
--- a/nix/nixos-modules/users/root.nix
+++ b/nix/nixos-modules/users/root.nix
@@ -17,6 +17,10 @@ in
   options.scale-network.users.root.enable = mkEnableOption "user root and sudo configs";
 
   config = mkIf cfg.enable {
+
+    security.pam.sshAgentAuth.enable = true;
+    security.pam.services.sudo.sshAgentAuth = true;
+
     security.sudo = {
       extraConfig = ''
 #        Defaults rootpw
--- a/nix/nixos-modules/users/root.nix
+++ b/nix/nixos-modules/users/root.nix
@@ -25,7 +25,6 @@ in
       extraConfig = ''
 #        Defaults rootpw
         Defaults lecture="never"
-        Defaults env_keep+=SSH_AUTH_SOCK
       '';
     };
 
--- a/nix/nixos-modules/users/root.nix
+++ b/nix/nixos-modules/users/root.nix
@@ -17,6 +17,10 @@ in
   options.scale-network.users.root.enable = mkEnableOption "user root and sudo configs";
 
   config = mkIf cfg.enable {
+
+    security.pam.rssh.enable = true;
+    security.pam.services.sudo.rssh = true;
+
     security.sudo = {
       extraConfig = ''
         Defaults rootpw
--- a/nix/nixos-modules/users/root.nix
+++ b/nix/nixos-modules/users/root.nix
@@ -25,7 +25,7 @@ in
       extraConfig = ''
 #        Defaults rootpw
         Defaults lecture="never"
-	Defaults env_keep+=SSH_AUTH_SOCK
+        Defaults env_keep+=SSH_AUTH_SOCK
       '';
     };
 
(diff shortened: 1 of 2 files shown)


owendelong:
UPdated branch to include the rash library instead of the ssh_auth library. The latter is apparently no longer developed and does not support sk style keys. The rash library is newer, written in rust (not sure that's a feature), and includes support for sk style keys (tokenized keys that require interaction with a hardware token)

nix/nixos-modules/users/root.nix
```diff
@@ -17,10 +17,15 @@ in
...
+
+    security.pam.rssh.enable = true;
+    security.pam.services.sudo.rssh = true;
+
     security.sudo = {
       extraConfig = ''
-        Defaults rootpw
+#        Defaults rootpw
         Defaults lecture="never"
+	Defaults env_keep+=SSH_AUTH_SOCK
```
MatthewCroughan: You probably shouldn't need to do use env_keep here, as this is kept for you by this logic here, so you should be able to remove this extraConfig.

https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/modules/security/pam.nix#L1777-L1782

nix/nixos-modules/users/root.nix
```diff
@@ -17,9 +17,13 @@ in
...
 
   config = mkIf cfg.enable {
+
+    security.pam.rssh.enable = true;
+    security.pam.services.sudo.rssh = true;
+
     security.sudo = {
       extraConfig = ''
-        Defaults rootpw
+#        Defaults rootpw
```
MatthewCroughan: Any reason to keep this comment in?

owendelong: It's gone now. Was hesitant to remove it at first in case we decided to revert this for some reason, but I'm done fighting with the stupid annoying nixfmt asshat shit.

MatthewCroughan (approved):

kylerisse:
I don't understand why we want or need this. I prefer the existing method of 2 factor auth for root access.

nix/nixos-modules/users/root.nix
```diff
@@ -17,9 +17,12 @@ in
   options.scale-network.users.root.enable = mkEnableOption "user root and sudo configs";
 
   config = mkIf cfg.enable {
+
+    security.pam.rssh.enable = true;
+    security.pam.services.sudo.rssh = true;
+
     security.sudo = {
       extraConfig = ''
-        Defaults rootpw
```
MatthewCroughan, line 22: You should probably keep this in, so that users not using an ssh agent can still use the expected password.

MatthewCroughan (changes requested):

nix/nixos-modules/users/root.nix
sarcasticadmin: I agree, we should leave this as a fallback

nix/nixos-modules/users/root.nix
sarcasticadmin: `nix fmt`

sarcasticadmin (changes requested):
thanks for testing this on dev-server

I would agree with @MatthewCroughan we should not be assuming everyone has ssh-agent enabled and leave sudo password for root enabled.

nix/nixos-modules/users/root.nix
owendelong: nix fmt can kiss my asterisk.

owendelong: And the rootpw thing is back.

owendelong:
> I don't understand why we want or need this. I prefer the existing method of 2 factor auth for root access.
Then don't use it -- Nothing says you have to. The rootpw option is still available to you.

kylerisse:
> > I don't understand why we want or need this. I prefer the existing method of 2 factor auth for root access.
Then don't use it -- Nothing says you have to. The rootpw option is still available to you.

great, looks good, it wasn't when I added the comment

sarcasticadmin (approved):
`confirmed` working on the `dev-server` while keeping the root password. Testing with my yubikey (ed25519 key + gpg-agent):

- plug in key -> enter pin -> ssh user@dev-server -> sudo -> root shell (Works)
- plug in key -> enter pin -> ssh user@dev-server -> remove key -> sudo -> root pw prompt -> root shell (Works)

PR #1188, opened 2026-04-29
Add smartmontools package to standard build

owendelong:
## Description of PR

Adds smartmontools (smartctl, etc.) to the standard package set for nix installations

 ## Previous Behavior

smartctl: Command not found

## New Behavior

smartctl gives useful data on disk behavior and history

## Tests

surgical tubing, lubricants, and a yak.


sarcasticadmin (approved):
I agree, it's a sensible thing to have available in each machine :shipit:

